PR #7, opened 2017-11-09
See if travis has geos installed, and if so whether that was interfering.

ceball:
I couldn't leave it alone, Philipp, even though I wanted to...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ language: generic
 
 os:
   - linux
+  - osx
 
 # get a more recent version of osx than the default image
 osx_image: xcode8.3
@@ -14,6 +15,11 @@ sudo: false
 
 env:
   - PYTHON_VERSION="3.6"
+
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+      brew uninstall geos;
+    fi
 
 install:
   # Install conda
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,14 +4,13 @@
 # https://github.com/travis-ci/travis-ci/issues/4729).
 language: generic
 
-os:
-  - linux
-  - osx
 
-# get a more recent version of osx than the default image
-osx_image: xcode8.3
-
-sudo: false
+matrix:
+  include:
+    - os: linux
+    - os: osx
+      language: generic # python doesn't seem to work on osx images
+      if: branch = master AND type != pull_request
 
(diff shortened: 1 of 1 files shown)


philippjfr:
Wow, so the machines did actually have brew installed gdal and libgeos on them! That explains a lot.

```
6.53s$ if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew uninstall --ignore-dependencies geos gdal postgis; fi
Uninstalling /usr/local/Cellar/geos/3.6.2... (402 files, 8.6MB)
Uninstalling /usr/local/Cellar/gdal/1.11.5_3... (232 files, 30.7MB)
Uninstalling /usr/local/Cellar/postgis/2.4.0... (228 files, 91MB)
```

ceball:
Seems so.

I wish there were integration between conda and travis rather than all these other package managers. Still, this kind of thing does help catch real world scenarios that will come up for users (and shows problematic packaging too...).

ceball:
Looks like that fixed it? I cancelled the linux builds early on, which is why the status is red I think.

I propose merging this at the same time as I fix #8.

jbednar:
Do you think these are coming from osx_image: xcode8.3?  I.e. for some reason switching to that is installing a bunch of other libraries?

ceball:
Brew-installed stuff is present in every image, I think. Versions of things will presumably vary between images, which might affect exactly what interference occurs. There are some indications of what kinds of things are installed in the images at https://docs.travis-ci.com/user/reference/osx/. A few weeks ago, 8.3 became the default on travis.

It'll be like this out in the real world on people's own machines, too, so these "travis" failures are useful, I think, even though they are unbelievably annoying to us. The real thing in my opinion is that the conda package(s) shouldn't be finding 'system' libraries in preference to ones supplied via conda. I'm not certain exactly what is happening or which packages are involved; the packages that look like they might be involved here are all outside my area of experience.

ceball:
It would also be very nice if conda were integrated with travis in some way, as an alternative to other package managers such as homebrew. Apart from anything, many projects spend a lot of time just waiting on conda env create. 

(I've noticed it's usually faster to download the 0.5 GB full Anaconda distribution every time and add a few packages into it than it is to download the tiny miniconda and have to add everything you need.)

ceball:
I'll merge and we'll see how it goes. Travis has been having more osx trouble recently (no network access)...

PR #12, opened 2017-11-10
Consolidated apps directories

jbednar:
Having two apps directories was confusing.  Merged them into one directory, at the top level.  The top level makes more sense semantically, so that one can choose between apps or notebooks to work with; apps don't need to have any relation to notebooks at all.  The only downside I can see is that notebook 13 had a link to edit a file from an app using Jupyter's internal editor, and I had to remove that link because it doesn't seem to allow editing files outside of the directory where Jupyter was launched.  If that can be worked around, please restore that line!  Or feel free to discard this PR and make a different one using notebooks/apps, if having that link seems very important.


philippjfr:
Any idea what's up with the tests? Don't have permissions to see the output.

philippjfr (commented):
Please merge it if you can figure out the tests. I'm happy with it.

jbednar:
There isn't any output:

image

jbednar:
Why don't you have access to the anacondaviz tests?  You're an admin on the project.  Do you mean, access at appveyor?  I'm not even sure why only appveyor is in the list, and not Travis.  I don't think the tests are set up yet, maybe?

jbednar:
I guess I'll merge; this just seems untested rather than failing.

philippjfr:
Tests were set up but I think they broke when I made the repo public.

jbednar:
Ah, could be!

PR #14, opened 2017-11-13
02df

jbednar:
Rewrote tutorial 02 to start with dataframes.  Considered replacing it with http://holoviews.org/getting_started/Introduction.html , which already does that and perhaps is more compelling at the start, but it seems like the additional material covered here would be hard to duplicate.


jbednar:
Closed; will rebase on new master.

PR #21, opened 2017-11-28
Renamed anacondaviz to pyviz in site files

jbednar:


ceball:
Merging as can't be any less broken than master

PR #22, opened 2017-11-28
WIP: website

ceball:
https://pyviz.github.io/pyviz/ is built on pushing to this PR.

To do:

* [x] Fill in placeholder content, making coherent

Before merging:

* [ ] squash
* [x] update travis file to e.g. build site on tag only


jbednar:
In the commits above, I removed the links to the individual tutorial sections from the main tutorial page, because they had only placeholder text describing them, whereas the 00 Welcome notebook already has full descriptions of each one.  Can we get rid of tutorial/index.rst and simply have it point to tutorial/00_Welcome.html ?  That way users of the website will see the same starting point as users of the notebooks, both of them showing the same content.

Also, right now none of the links to the exercises from 00_Welcome work, because it doesn't look like the exercises are built.  I'm not sure where to add those?

I've added those two tasks to the to-do list above.

PR #28, opened 2018-02-19
Initial version of pyviz module

jbednar:
PR made here for discussion.  At the moment, there isn't anything substantial in this `pyviz` module; all it does is make the various PyViz packages available as short aliases to simplify public examples and provide a more concise API for people who want to treat PyViz as a single unit with subpackages.  E.g. instead of:

```
import holoviews.operation.datashader as hd
import holoviews as hv
hv.extension("bokeh")
hd.dynspread(hd.datashade(hv.Curve([1,10,100])))
```

it can be 

```
from pyviz import hv, hd
hv.extension("bokeh")
hd.dynspread(hd.datashade(hv.Curve([1,10,100])))
```

This approach is most important for small, compact examples that involve multiple packages, which is what PyViz is all about.  For much larger examples, it's ok to use the more-verbose but more specific individual exports, and for single packages it's not an issue.

Plus, once there is a conda package, people can install using `conda install -c pyviz pyviz` instead of downloading an environment file, setting up an environment, and activating it.  Having this module doesn't stop anyone from importing only whichever little bits they need; it just provides a one-stop-shopping alternative.

For discussion:

1. Is it reasonable to conditionally import geoviews, as here, in case someone's environment drifts and e.g. they get some incompatible GDAL library involved and break a gv installation they aren't even using?  gv would always be installed with pyviz, so this may not be important.
2. Should it import matplotlib (as mpl?)?
3. Should it import xarray (as xr)?  xarray is already required by datashader, so I guess it won't hurt.
4. How do people feel about `pm` as an abbreviation for `param`?
5. What about `parambokeh`? `pb`?
6. Any other concerns?


philippjfr:
>make the various PyViz packages available as short aliases to simplify public examples and provide a more concise API for people who want to treat PyViz as a single unit with subpackages

I can't say I'm a huge fan, I wouldn't expect users to import PyViz in their day to day work and if we start using these aliases in examples it obscures where a user would import these from in practice.

>Is it reasonable to conditionally import geoviews, as here.

Running all these imports for all examples even when they are not needed will add significantly to import time and make the examples feel sluggish. We're trying to get away from importing everything in HoloViews and this approach compounds the problem.

So overall I'm not really convinced this is a good idea, but if you insist those abbreviations sound fine.

ceball:
I am really sympathetic to this PR. Anything to shrink the big chunk of intimidating 

```
import module1 as m1
import module2 as m2
import module3 as m3
import module4 as m4
import module5 as m5
```

text plastered at the start of every example, while not "destroying namespaces", seems like a great idea to me. I think it's really worth trying to make the pyviz examples as simple as possible for people unfamiliar with it all (python, these packages), if it can be done without obscuring too much what things really are. 

When someone first starts, they probably do not care what all the packages are, they just want to know what can be done and how difficult it is to do. I think every extra line and word make it less likely they'll proceed. And it's not deceptive to simplify the imports as suggested in this PR - it's only one step to discover what the packages really are.

At the same time, I share some concern about `from pyviz import m1, m2, m3, m4, m5` being different from how a user would import m1, m2, m3, m4, m5 outside of a pyviz example. And I also worry that people would get confused by there being "another module" (i.e. pyviz). "Wait, am I supposed to import from pyviz, or just import the package?" kind of thing. Also, some people do care a lot about the imports, and study those first to know what they are in for when they read the rest (I'm one of them...), so they might be annoyed to have to investigate. But what with all the logos and pink warnings that get printed all the time, it's probably easy to identify the packages anyway ;)

> Running all these imports for all examples even when they are not needed will added significantly to import time and make the examples feel sluggish. We're trying to get away from importing everything in HoloViews and this approach compounds the problem.

Would it be a horrible idea to do some kind of lazy importing, to avoid this problem?

On the other hand, many examples already import lots of these things anyway, so they will already incur the cost of the imports - meaning maybe we shouldn't worry about that here? And the packages themselves ought to sort out their long import times?

philippjfr:
I agree that the huge block of imports many of our examples have are scary so I should emphasize that I too would like a good solution to this issue. I also don't yet have a better suggestion.

>Would it be a horrible idea to do some kind of lazy importing, to avoid this problem?

Seems sensible to me if it can be done in a nice way.

>And the packages themselves ought to sort out their long import times?

That's definitely a good idea but the question is if it will be enough. I think the data interfaces in HoloViews should have some mechanism for lazy loading and there's probably some other things we can do. Would have to run some detailed benchmarks but my main worry is GeoViews, which has to import a number of heavy weight dependencies. To make this concrete on my system takes 1.7 seconds to import holoviews and geoviews takes 4 seconds. So at least as of right now, using PyViz for pure holoviews examples doesn't seem like a good idea, as that's a huge hit to take.

jbednar:
Just to be clear, I'm not proposing using `pyviz` for *any* examples in the individual projects.  For one thing, pyviz isn't and presumably couldn't be a dependency of those projects, as *they* are dependencies of pyviz.  I also don't think any PyViz project should assume the user has installed `pyviz`; from the perspective of any single project, `pyviz` doesn't add anything of value.  

I'm only proposing that there be an entry point where the examples are simpler and don't immediately *force* users to figure out which package is which, as long as they cut and paste properly.  Yes, eventually they will need to figure out the packages.  But that shouldn't be the majority of what they see in our simple examples that are meant to ease people into the PyViz approach.

The loading time seems like an orthogonal issue, for several reasons: (1) for me, `time python -c 'import pyviz'` reports 3.4 seconds, which is annoying and should be improved, but isn't prohibitive. (2) Our examples eventually import all this stuff anyway, because for concreteness we tend to use geographic examples. (3) If we do set up [lazy loading](https://github.com/mnmelo/lazy_import/blob/master/lazy_import/__init__.py), seems like we could do that as an optimization, regardless of this PR, and presumably only for datashader and geoviews, as the rest are required for just about any example anyway.

philippjfr:
>I'm only proposing that there be an entry point where the examples are simpler and don't immediately force users to figure out which packages is which, as long as they cut and paste properly

So are you suggesting updating all the tutorials which do not use GeoViews or Datashader to import from PyViz? I'm fine with the idea if we introduce the alias in some early tutorial and then use it in simple examples. It'll also motivate me to improve the import times.

jbednar:
I'm proposing using it in examples, more than in tutorials; short snippets that do something the user can appreciate.  I guess these would form the PyViz gallery?

jlstevens:
Here is the list of abbreviations that I think are being proposed

```python
from pyviz import gv, hv, hd, ds, bk, tf, pm, pb, xr
```

Any more? I think I'm ok with this though I am not a huge fan of ``bk`` and I'm not sure about ``pm``. I'm also not sure if this is easier for people or just more confusing....

philippjfr:
Might as well add numpy and pandas right?

jlstevens:
So this then?

```python
from pyviz import gv, hv, hd, ds, bk, tf, pm, pb, xr, np, pd
```

jbednar:
Yes, that's the full list of what's in there already.

ceball:
The updated dependencies are causing a problem on windows (appveyor, and I can now reproduce locally):

screen shot 2018-02-20 at 10 12 49 pm

This does not happen on master, so I guess it's to do with either the bokeh or datashader upgrade. (Note to selves: there are too many channels in this environment for it to be practical...)

With conda<4.4, the error message about the holoviews package goes away, but the others remain.

ceball:
I believe the problem on windows is from `packaging` on the bokeh channel:

screen shot 2018-02-20 at 10 19 48 pm

ceball:
The holoviews and hdf5 "errors" from conda are not actually fatal:

screen shot 2018-02-20 at 10 43 20 pm

I couldn't tell for sure initially when they were all present, but it's only the packaging one that was fatal.

jbednar:
That's very enlightened;; indeed, it is very unlikely to matter that the COPYING and RELEASE.txt files are clobbering each other.  Seems like something to report to the maintainers of those conda packages.  So now that packaging is installed from a different channel it's ok?

jbednar:
So if the tests pass on windows, it's ready to merge?

ceball:
Yes, I think so (from the point of view of things working). I guess everyone's happy enough that we try out the pyviz module as in this PR? I certainly can't think of a better alternative right now.

ceball:
> That's very enlightened;; indeed, it is very unlikely to matter that the COPYING and RELEASE.txt files are clobbering each other

I think in a way it's good that conda informs about this. However, it could potentially be important clobbering going on, so the situation is a bit scary. Messages like these will generally just fly by and be missed or ignored (I only noticed them because they were immediately followed by a fatal error).

PR #29, opened 2018-02-20
Pin versions of all dependencies (direct or not) 

ceball:
To versions known to work well together for running the examples.


jbednar:
Replaced with pinning only the packages expected to diverge over time; should separately provide an option for pinning all packages.

PR #30, opened 2018-02-20
WIP: conda package and pyviz commands to get data etc

ceball:
* [ ] Express required dependencies in setup.py (`install_requires`).
* [ ] Express test dependencies in setup.py (`tests_require`).
* [ ] Express dependencies for examples in setup.py (`extras_require['examples']`).
* [ ] Express dependencies for doc building in setup.py (`extras_require['docs']`). Note: I would expect this to just be the dependencies for the examples, plus `nbsite`.)
* [ ] Automatically generate pyviz conda package(s) with (required? all? combinations of?) dependencies.
* [ ] Automatically generate pyviz environments (default/minimal and "fully pinned")?
* [ ] Switch to using autover for consistent versioning.
* [ ] Add `pyviz` command with actions including `download_sample_data` and `download_examples`.

Note 1: I am moving bit by bit from the existing state to that proposed above, so the code is not yet anywhere near ready to review.

Note 2: I'm hoping the download functions can be shared across projects, but bundled with each project at packaging time, so there's no dependency on a cross project download data/examples library.


jbednar:
That *all* sounds great!!  You might find https://github.com/gbrener/python-utils/blob/master/import_utils.py useful for collecting what the examples depend on.  

The only bit that's not clear is how many packages.  My guess is `pyviz`, with minimal dependencies (those in the module in the other PR), and `pyviz-examples`, which would depend on pyviz to get the main dependencies but would add any that are used in the examples.  Sound ok?

conda.recipe/meta.yaml
```diff
@@ -37,12 +37,11 @@ test:
     {% endfor %}
   # WIP: will replace with commands for fetching stuff etc.
   source_files:
-    - download_sample_data.py
-    - datasets.yml
     - etc/test_notebooks.py
     - notebooks/
   commands:
-    - python download_sample_data.py
+    - pyviz download_examples
```
jbednar: Shouldn't this be `install_examples`?  Seems like the version of the examples to use should be determined by the version of this module, rather than whatever is on some external site, and so the task is to extract the examples from wherever they are on the filesystem after the package was installed, and install them in the user's directory so they can be run and modified, as for the datashader examples.

ceball: In the datashader packages (at least the few I looked at), examples/ is in site-packages. When installed, examples/ is written into site-packages as a package. That seems worse than asking someone to download the examples.

```
(earthsim) [180223 173505]~$ python
Python 3.5.5 | packaged by conda-forge | (default, Feb 13 2018, 05:40:31) 
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import examples
>>> examples.__file__
'/Users/cball/anaconda3/envs/earthsim/lib/python3.5/site-packages/examples/__init__.py'
>>> 
(earthsim) [180223 173516]~$ ls /Users/cball/anaconda3/envs/earthsim/lib/python3.5/site-packages/examples/
README.md			environment.yml			nyc_taxi-nongeo.ipynb		raster.py
__init__.py			filetimes.py			nyc_taxi-paramnb.ipynb		solar.ipynb
__pycache__			filetimes.sh			nyc_taxi.ipynb			streaming-aggregation.ipynb
census-congressional.ipynb	filetimes.yml			opensky.ipynb			streaming.py
census-hv-dask.ipynb		get_raster_data.sh		osm-1billion.ipynb		taxi_preprocessing_example.py
census-hv.ipynb			getting_started			osm.ipynb			topics
census.ipynb			holoviews_datashader.ipynb	packet_capture_graph.ipynb	trajectory.ipynb
dashboard			hover_layer.ipynb		pcap_to_parquet.py		tseries.ipynb
datasets.yml			landsat.ipynb			pipeline.ipynb			user_guide
download_sample_data.py		legends.ipynb			plotting_pitfalls.ipynb
edge_bundling.ipynb		lidar.ipynb			race_elevation.ipynb
(earthsim) [180223 173519]~$ head /Users/cball/anaconda3/envs/earthsim/lib/python3.5/site-packages/examples/README.md 
# Datashader Examples

The best way to understand how Datashader works is to try out our
extensive set of examples. Static versions of most of them are
provided on [Anaconda Cloud](https://anaconda.org/jbednar/notebooks),
but for the full experience with dynamic updating you will need to
install them on a live server. To get started, first go to your home
directory and download the current list of everything needed for the
examples:
```

However, I agree that the examples could be part of a package, and then installed by the user for editing etc. I think we suggested a pyviz-examples package. I haven't got that far yet.

Meanwhile, the download command is useful for testing as I transition from a git clone + environment to one or more packages, where the installed packages actually get tested. The examples aren't in any of the packages yet, so downloading is easier. I have control over the notebooks and apps that are downloaded (e.g. I can have just a few cut-down ones for faster testing iterations) without needing to mess with committing lots of changes to notebooks etc. So I'd prefer to continue without changing this for now, but if it's a problem I can change it now.

jbednar: Ideally, the examples would be installed in share/, not site-packages/, but I believe they ended up in site-packages/ because of difficulties ensuring that the examples are available on various platforms and installation methods, and also difficulties in finding them once installed.  I'd be happy to see them move to share/{projname}/examples if that can be done reliably.

In any case, to me it still seems vastly preferable for them to be installed somewhere alongside the code, even if that means it has to be in site-packages/.  That way we can be sure that people get the right version of the examples, which is crucial because the examples are always the way that we communicate any new features, and thus any particular version of the examples will never work with any older version of the code.  If we provide the examples in any separate form, it seems unlikely that we will be able to have a matching set, particularly in between releases (as in other cases you support having packages available continuously between releases). We could of course store all versions of the examples in some sort of lookup table by version, but that seems messy and error prone to me.

A less important but still relevant criterion is that if the examples are installed with the code, they will be available even in airgapped or mobile environments that can't access an online site.  The total size of the examples (since they don't have output included) seems small enough that it shouldn't matter whether they are shipped or downloaded, unlike the example data (which is prohibitively large, but at least never changes with new versions except to add new files).

jbednar: Note that `pyviz-examples` was about installing *dependencies*; it doesn't change anything about what happens to the examples themselves.  If the examples are included in `pyviz-examples`, they still have to go somewhere on the file system (share/ or site-packages/ or wherever), they still need to be findable reliably, and they still need to match the version of the code installed (which is now more difficult if the examples are in a different package).

jbednar: Anyway, you can do what you like for development and debugging, but I suggested renaming that target because I expected it to be focused on installing the examples, regardless of whether they come from a local or remote source, and so it shouldn't be named "download". Hopefully the users won't have to care where they come from, as long as they get the right version, but they do care that the result of the operation will be that they have a copy of the examples ready for them to edit.

jbednar:
Closed for now as it has diverged from the current setup; can reopen or replace with a new PR.

PR #35, opened 2018-03-05
Update Travis configuration to match reorganized code

jbednar:
When the pyviz module commands for installing examples and downloading data were added, they broke the travis configuration because the underlying scripts moved into the pyviz directory.  This PR should address those breakages while also substituting a smaller test file for normal testing to avoid downloading the entire taxi dataset.


jbednar:
This should be squash-merged as soon as the tests pass, to make master start passing.  

@philipp, I updated the meta.yml dependencies to match what was needed to get the tests to run, though the tests are using environment.yml rather than meta.yml.  Can you please skim the diffs and if it looks good to you can you rerelease the package? E.g. we had removed scikit-image since it's not referred to in the tutorials, but I had to put it back because it's actually used internally by the datashader graph support.  Not sure why geopandas wasn't in the list before, but we definitely do need that for the tutorials as well.

PR #39, opened 2018-03-06
Create pyviz.exe on windows. 

ceball:
Also, added a test that the pyviz command runs.

The conda recipe needs to be cleaned up, but this should fix the pyviz command on windows for now.

screen shot 2018-03-06 at 12 15 00 pm


jbednar:
Thanks! I went ahead and merged it, but will check tests on master just in case. Do we need a new release of the package before this will have an effect?

PR #41, opened 2018-03-06
Replaced 01 Workflow Introduction

jbednar:
The previous workflow introduction focused on widgets and datashader, whereas this more recent material is broader based and gives a better overall introduction.  It's an entirely different notebook and should not be compared to the previous one line by line.


philippjfr:
Since the travis builds passed and the windows builds are going to be queued for a while I'm just going to merge.

PR #44, opened 2018-03-09
Upgrade nbsite

ceball:
Switch to nbsite and sphinx_ioam_theme dev conda packages (rather than conda install deps + pip install nbsite and sphinx_ioam_theme), with the side effect of getting the 'title fix' included in the latest sphinx_ioam_theme.

Also, make latest build of docs (for any branch except master) visible at https://ioam-docs.github.io/pyviz-latest/. (Latest build of master already visible at https://ioam-docs.github.io/pyviz-master/)

Note: this PR also contains a fix for travis/gh pages deployment failing (see https://github.com/travis-ci/travis-ci/issues/9312).

To do before merge:
  * [x] check title issue is fixed at https://ioam-docs.github.io/pyviz-latest/ when build completes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,6 +71,7 @@ jobs:
             tags: true
         - edge:
             branch: v1.8.47
+          provider: pages
           skip_cleanup: true
           github_token: $GITHUB_TOKEN
           local_dir: ./doc/_build/html
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,9 @@ jobs:
           github_token: $GITHUB_TOKEN
           local_dir: ./doc/_build/html
           repo: ioam-docs/pyviz-latest
+          on:
+            all_branches: true
+            condition: $TRAVIS_BRANCH != master
 
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,20 +61,25 @@ jobs:
         - nbsite_cleandisthtml.py ./_build/html take_a_chance
         - cd ..
       deploy:
-        - provider: pages
+        - edge:
+            branch: v1.8.47
+          provider: pages
           skip_cleanup: true
           github_token: $GITHUB_TOKEN
           local_dir: ./doc/_build/html
           on:
             tags: true
-        - provider: pages
+        - edge:
+            branch: v1.8.47
           skip_cleanup: true
           github_token: $GITHUB_TOKEN
           local_dir: ./doc/_build/html
           repo: ioam-docs/pyviz-master
           on:
             branch: master
-        - provider: pages
+        - edge:
+            branch: v1.8.47
+          provider: pages
           skip_cleanup: true
           github_token: $GITHUB_TOKEN
           local_dir: ./doc/_build/html
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,6 +74,11 @@ jobs:
           repo: ioam-docs/pyviz-master
           on:
             branch: master
+        - provider: pages
+          skip_cleanup: true
+          github_token: $GITHUB_TOKEN
+          local_dir: ./doc/_build/html
+          repo: ioam-docs/pyviz-latest
 
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,9 +50,7 @@ jobs:
       stage: doc
       script:
         - python pyviz/download_sample_data.py data
-        - conda install -c conda-forge 'sphinx<1.7' beautifulsoup4 graphviz
-        - pip install nbsite
-        - pip install sphinx_ioam_theme
+        - conda install -c pyviz/label/dev -c conda-forge nbsite sphinx_ioam_theme
         - cd doc/tutorial
         - nbsite_nbpagebuild.py pyviz pyviz ../../notebooks . 1
         - cd ..
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -8,9 +8,8 @@ authors = u'PyViz authors'
 copyright = u'\u00a9 2017-2018, ' + authors
 description = 'How to solve visualization problems with Python tools.'
 
-# TODO: gah, version
-version = '0.0.1'
-release = '0.0.1'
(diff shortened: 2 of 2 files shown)


ceball:
@jbednar, does this fix the title issue you found? You can see the output at https://ioam-docs.github.io/pyviz-latest/ (assuming no other build overwrites it before you look...)

ceball:
Note: there's no title on the index page. However, that's not related to this issue (it was true before and is still true). It's more related to nbsite not supporting index.ipynb yet (https://github.com/pyviz/nbsite/issues/23). Meanwhile, we should add a title to the index page. Currently it seems to be done like this:

```
.. raw:: html

   <p style="font-size:30px"><b>How to solve visualization problems with Python tools</b></p>
```

I.e. not in a way sphinx can use. So I'll just figure out how to add the same thing as the page title for now.

ceball:
When the latest build finishes, https://ioam-docs.github.io/pyviz-latest/ should show reasonable titles for every page, including the index page. The index page title will improve in the future via https://github.com/pyviz/nbsite/issues/23.

ceball:
Still shows "no title" for the index page. I'll have to try again...

jbednar:
> I.e. not in a way sphinx can use. So I'll just figure out how to add the same thing as the page title for now.

The title for the home page should be the name of the project, for this project and probably for all projects by default.  I apologize for how the page title for the home page is done right now; if you can figure out a better way to get that page to format correctly with all the logos on it, please do!

.travis.yml
```diff
@@ -63,19 +61,33 @@ jobs:
         - nbsite_cleandisthtml.py ./_build/html take_a_chance
         - cd ..
       deploy:
-        - provider: pages
+        - edge:
+            branch: v1.8.47
```
jbednar, line 65: For my own enlightenment, can you explain what this setting does?  It's very mysterious. :-)

philippjfr:
The titles on http://geoviews.org/ and http://holoviews.org/ seem okay and at least geoviews.org is using nbsite. Any idea why that would differ?

jbednar (approved):
Looks great, thanks!  Should be merged as soon as the index page can have a decent title.

.travis.yml
philippjfr: It downgrades some travis component because the most recent version is broken: https://github.com/travis-ci/travis-ci/issues/9312

ceball:
> The titles on http://geoviews.org/ and http://holoviews.org/ seem okay and at least geoviews.org is using nbsite. Any idea why that would differ?

If you're asking about no title on the pyviz homepage: as noted above there's no title defined in pyviz's index page (it's instead done with formatting in html). Presumably in other project there is a title for the index page. Anyway, I hope that when the latest doc build finishes, there'll be a reasonable default title for the case a page has no title. (The better eventual solution will be replacing index.rst with index.ipynb in general: https://github.com/pyviz/nbsite/issues/23) 

If you're asking about markup that was appearing after the dash in page titles for pyviz: that's because in pyviz's conf.py, the project name is defined as `project = u"<b><span style='color:#4792cf;'>Py</span><span style='color:#fdca48;'>Viz</span></b>"`. Presumably the project name is just a string with no tags in the other projects. I've already fixed this problem by stripping tags for the html page title.

philippjfr:
Thanks for the clarification, makes sense. Was just trying to make sure that any fix in nbsite wouldn't end up breaking the titles for other projects.

ceball:
The changes are in sphinx_ioam_theme, and shouldn't have an effect on other projects:

First change:
`<title>{{ title }} &mdash; {{ shorttitle }}</title>` is now `<title>{{ title|striptags|e }} &mdash; {{ shorttitle|striptags|e }}</title>` (i.e. stripped of tags and escaped)

Normally the html page title looks like `Title - Project 0.1.2 documentation`.

The other change is that if there's no title (rare - only pyviz index page?), the title will just look like `Project 0.1.2 documentation`.

.travis.yml
jbednar: Ah; very appropriately mysterious, then.  Should probably open an issue somewhere to remind us to remove this setting once Travis fixes the issue, as it will no doubt eventually cause us problems.

jbednar:
> If you're asking about markup that was appearing after the dash in page titles for pyviz: that's because in pyviz's conf.py, the project name is defined as `project = u"<b><span style='color:#4792cf;'>Py</span><span style='color:#fdca48;'>Viz</span></b>"`. Presumably the project name is just a string with no tags in the other projects. I've already fixed this problem by stripping tags for the html page title.

Sorry for causing that problem.  Is there/can there be a better place to put a wordmark for the project that will appear in the page banner?  I put that markup into the project name, but really it should be in something that only applies to the page banner, leaving the project name as bare text for use in other contexts where it appears.

jbednar:
> Normally the html page title looks like Title - Project 0.1.2 documentation.

Does the word "documentation" do anything useful here?  The title mainly shows up in the tab, where that bit is just cut off in Chrome.  Moreover, not every page in our project sites is really "documentation" in any strict sense, as many are discussion, background, etc.  I'd vote for removing that word; it would only apply to some designated subset of a site, and if we don't do that, I don't think we should falsely and redundantly claim everything in the project site is documentation.

> The other change is that if there's no title (rare - only pyviz index page?), the title will just look like Project 0.1.2 documentation.

I think that's reasonable, but "documentation" is even less appropriate on the main site index page.  Actually, can we force the main site index page always just to have the project name as its title (maybe without the version number?)  The way I've set up pyviz.org is how I eventually want all the sites to be, i.e. to have a logo with word mark in large size as the first thing on the page, which means that any subsequent heading is not likely to be useful as the page title. E.g. on pyviz.org the first heading is the tag line for the project, *not* the title of the page, because the title is now an image. I think that should be the standard on all of our projects that have logos, and to make the behavior consistent with projects that have no logo, it seems like the main index page should always just have the project name as its title by default. Sound ok?

ceball:
I've merged, because although it's not perfect, it's an improvement. I've linked to comments in this PR from relevant nbsite issues.

PR #45, opened 2018-03-09
Updated/simplified streaming tutorial

jbednar:
- Made it ok to execute cells out of order by making variables unique
- Left at least one plot running for "Run All" so that it's clear that streaming works without requiring close study
- Avoided clearing output to make the plots be easier to skim through and re-run without them all looking the same (empty).
- Simplified code and made it clearer which bits were about setting up a stream, and which bits are about streaming viz

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,7 @@ jobs:
     - <<: *default
       stage: doc
       script:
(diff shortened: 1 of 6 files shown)


jbednar:
Not sure why the 0.9.8 commits are showing up here, but presumably they won't hurt?

PR #58, opened 2018-05-25
Fix fiona/poppler conflict in conda package (allows 07_Geographic_Data.ipynb to run)

ceball:
If poppler's installed, version 0.65.1 (at least) causes a problem for fiona on linux and mac (linux: https://travis-ci.org/pyviz/pyviz/jobs/383650789#L2372).

Not sure what is pulling in poppler (edit: installing fiona pulls in poppler, although it's not a direct dependency), but poppler's installed if you create the env or install the current released conda package.

poppler doesn't exist for windows. 

In the conda recipe I've pinned poppler to the last known compatible version (but not on windows).

For the environment, on the other hand, I don't know another solution except having a separate environment file. If we go with this, would need to update any installation instructions that mention it.

* [x] Once figured out, also update conda recipe
* [x] test the updated conda recipe and package
* [x] (edit: not necessary) update installation instructions if necessary


jbednar:
That's a nightmare!  Can we just force poppler to be removed after the envt is built, then separately figure out how to remove it from our indirect dependencies?

ceball:
Fixing the environment on master is lower priority than fixing the conda package. So, I've removed the updated environment file from this PR (opening #59 to track), and here have just fixed the version of poppler on travis with a command after creating the environment.

ceball:
@jbednar, ready to merge from my point of view

ceball:
Although this works fine on linux and mac, it doesn't work on windows (where poppler does not exist). I *think* I tracked the issue down to https://github.com/conda/conda/issues/7344; will wait to see the outcome of that issue. I still think this PR could be merged to fix master.

Fixing the existing pyviz package for windows could be done in a number of ways, unrelated to this PR. The easiest way from our point of view would be to tell people to install pyviz >= 0.9.7-1. The easiest way for users would be if we were to upload 0.9.7-2 without run_constrained, and then upload 0.9.7-3 with run_constrained. Mac and linux users would then get 0.9.7-3 and the right version of poppler, while windows users would get 0.9.7-2 (no poppler).

The alternative is to fix the incompatibility between fiona and poppler, then get new c-f packages, then new packages on defaults. I'll at least report the problem, anyway, once I figure out where to do that.

ceball:
I've also filed https://github.com/AnacondaRecipes/aggregate/issues/60.

ceball:
> The easiest way for users would be if we were to upload 0.9.7-2 without run_constrained, and then upload 0.9.7-3 with run_constrained.

I've uploaded those package, and tested on mac, linux, and windows. win gets build 2, while linux and mac get build 3.

jbednar:
Can this be closed?  I didn't have any problems with notebook 07 today, and I vaguely remember the conda team saying that they were fixing something to do with poppler. Or does it need to be merged to make Windows work?  If so, @ceball, please merge it.

PR #64, opened 2018-06-15
Added SciPy 18 index

jbednar:
First, reorganized existing welcome and index pages to make it feasible to have multiple index pages, one for each presented tutorial (currently SciPy 2018).  Next, duplicated the index so that it can be customized for SciPy 2018.  It seems to build ok locally, but it's only rebuilding parts of it, so the real test will be when Travis builds it.


jbednar:
It looks like it mostly built fine (see https://ioam-docs.github.io/pyviz-latest), except what looks like a predictable consequence of https://github.com/ioam/parambokeh/pull/62 , i..e. that the custom index.rst file is overwritten by https://ioam-docs.github.io/pyviz-latest/tutorial/index.html .

@ceball, it looks a little tricky to get around that. For now, presumably the website-building code can simply revert the bogus auto-generated index.rst file and use the git-controlled one instead, or (better) not generate any index.rst file when there already is one.  

But it seems like the .toctree in the custom file ought to be auto-generated anyway; it doesn't seem to have any information in it that's not discoverable automatically.  I'd certainly rather not have to write index.rst by hand and have to update it every time I add a new tutorial section.  However, generating the rest of index.rst automatically seems problematic, because I had to add the section heading explicitly in the [index.rst](https://github.com/pyviz/pyviz/blob/9c419601d6466cbae04460095670be7910640dec/doc/tutorial/index.rst) or else the various tutorial sections showed up at the wrong level in the sidebar (at the top level, rather than grouped under Tutorial).  To be able to auto-generate that part of the index.rst file, it seems like one might have to pull out the title from the index.ipynb file and explicitly put that into the .rst file?  Tricky!

PR #93, opened 2018-09-11
Try to clarify examples overwriting.

ceball:
I tried not to make things too complicated. Maybe you can do better?


ceball:
One thing I just realized is that if tutorials have been removed between times someone's run install-examples, that could get confusing. But I think that's unlikely to be the case - right?

ceball:
Test failures are just that I cancelled the builds, since they don't matter (and there are ones after that do).

PR #99, opened 2018-09-13
Remove user- and system-specific metadata from notebooks

ceball:
I doubt this can be merged as it is right now - probably some important metadata has been removed. On the other hand, when I try to edit some of the notebooks, I get spurious diffs.

Can someone say what metadata should stay?

Don't suppose anyone now has enough experience of a tool that they can recommend it?

In the past, I listed some possible tools to consider at https://github.com/pyviz/pyct/issues/3, but never followed any of that up.

### DIFF
--- a/exercises/Exercise-1-making-data-visualizable.ipynb
+++ b/exercises/Exercise-1-making-data-visualizable.ipynb
@@ -48,9 +48,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "collapsed": true
-   },
+   "metadata": {},
    "outputs": [],
    "source": []
   },
@@ -142,9 +140,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "collapsed": true
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "dates = pd.date_range('2017-01-01', '2017-11-14')\n",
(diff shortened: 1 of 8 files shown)


jbednar:
All the metadata can be removed *except* from 01 and A2, which have per-cell metadata that controls how they are viewed using RISE slideshows.  So we can't merge the changes to 01 and A2, but the others should be fine. 

I'm usually using Jean-Luc's strip_kernel script, but it needs to be extended to preserve the RISE metadata, so I can't currently use it on 01 and A2 (and thus very often mistakenly strip out that metadata and have to laboriously recreate it!).  Jean-Luc also wants it to be implemented without jq, but I don't currently have that as a priority; just being safe to use on 01 and A2 would be plenty!

```
#!/bin/bash
cp "$1" /tmp/
/usr/local/bin/jq --indent 1 \
    '
    (.cells[] | select(has("outputs")) | .outputs) = []
    | (.cells[] | select(has("execution_count")) | .execution_count) = null
    | .metadata = {"language_info": {"name":"python", "pygments_lexer": "ipython3"}}
    | .cells[].metadata = {}
    ' "$1" > /tmp/stripped.ipynb
cat /tmp/stripped.ipynb > "$1"
```

ceball:
> I'm usually using Jean-Luc's strip_kernel script, but it needs to be extended to preserve the RISE metadata

I hope there's a viable alternative to that, but I'm out of touch. Having only python/jupyter dependencies (vs e.g. jq, bash, find, etc), available on pypi+anaconda.org, able to run concurrently, at least maintained somewhere visible/shared, etc etc all seem like reasonably important things...

jbednar:
Closed unmerged?

ceball:
I should have said when closing that there were conflicts in many
notebooks, which didn't seem worth sorting out vs just running the metadata
stripping tool again.

And the issue of having a tool+config to do this for pyviz in general is
already covered in the pyct issue mentioned above. Once that's done, travis
could run it on PRs and complain if there's a diff (assuming we can't do
anything fancier like have travis commit the cleanup).

jbednar:
Ok, thanks.

PR #104, opened 2018-09-23
Fix misc. doc typos

luzpaz:
Found via `codespell -q 3 -L te,ot,nd,ue`


philippjfr:
Thanks for these fixes, however fixing typos in the built docs won't make much of a difference since it will simply be rebuilt when the docs are rebuilt. The fixes have to be applied to the ipynb files.

ceball:
@luzpaz, thanks for trying :) It's unfortunate to miss out on these spelling fixes, so if you were able to figure out a way to do the spell checking on the source .ipynb files, that would be great...

PR #81, opened 2018-07-12
Improve packaging and testing

ceball:
Closes #5, #63, #67, #68

Changes:
  * dependencies, pins, and package metadata specified once, in setup.cfg (doesn't have to stay as setup.cfg forever, but having one non-code config file is a start...)
  * removed conda recipe; instead generated from info in setup.cfg
  * the environment.yml file has been moved to examples and is generated from info in setup.cfg
  * test matrix (envs+commands) specified once in tox.ini (nobody likes that file, but at least it's all in one place...)
  * reorganized examples to be like rest of pyviz projects
  * simplified nbsite usage (nbsite handles assets, has slightly improved interface, supports ipynb for more things)
  * can generate pip package (but: current pyviz depends on unreleased things e.g. bokeh 1.0 for which there are no pip packages)
  * converted readme to md
  * removed test_examples script (use nbsmoke)
  * replaced custom user examples/data commands with those from pyct (note: alters the commands)
  * fixed apps to run from any current directory
  * pyviz.org can be updated by `website` tag (or by a release tag)

Additions:
  * weekly "follow tutorial instructions" job (i.e. test released packages) on win, mac, linux
  * weekly "try without pins" job (i.e. see what upcoming problems are on the horizon, although with luck they might have passed by the time it comes for a new release)
  * dev website at ioam-docs.github.io/pyviz-dev (updated by dev tag or by `website_dev` tag).

To do:
  * [x] update docs
  * [x] request to enable cron on appveyor account, then configure it
  * [x] release new pyctdev [edit: there's a dev release with the new features, which should be enough for now]
  * [x] check the apps on windows (particularly the data paths: can the original, simpler path form be used?) [edit: data paths nothing to do with windows; the more complex form is needed unless people always run the app from a specified directory]
  * [x] test the pytest exclude command works for user install

Just before merge:
  * [x] restore package upload steps
  * [x] restore build step conditions

PR #105, opened 2018-09-24
Update instructions to use pyviz/label/dev.

ceball:
Currently, anyone following the installation instructions on pyviz.org will get an older pyviz (0.9.11), because the current package (0.9.12) depends on dev releases of several things. However, the tutorial on pyviz.org expects the latest pyviz (0.9.12).

I noticed this in #81, which includes testing the "user installation" of packages (and those tests are failing).

If merged, subsequently pushing a tag like website_update (on master) should update pyviz.org. (I think, from reading the travis file! In #81 the mechanism is changing to the standard pyviz scheme, where there's definitely a deliberate `website` tag to allow updates between releases.)

EDIT: I also updated the environment.yaml file - tests on master have been failing too, since they use the environment.

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -26,7 +26,7 @@ Step 2: Install ``pyviz`` and other required libraries
 
 ::
 
-   > conda install -c pyviz pyviz
+   > conda install -c pyviz/labe/dev pyviz
 
 
 Step 3: Install the tutorials in your current working directory and download data
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -26,7 +26,7 @@ Step 2: Install ``pyviz`` and other required libraries
 
 ::
 
-   > conda install -c pyviz/labe/dev pyviz
+   > conda install -c pyviz/label/dev pyviz
 
 
 Step 3: Install the tutorials in your current working directory and download data
--- a/examples/tutorial/00_Setup.ipynb
+++ b/examples/tutorial/00_Setup.ipynb
@@ -30,7 +30,7 @@
    "metadata": {},
    "source": [
     "```\n",
-    "! conda install -c pyviz pyviz\n",
+    "! conda install -c pyviz/label/dev pyviz\n",
     "! pyviz examples\n",
     "! cd pyviz-examples/tutorial\n",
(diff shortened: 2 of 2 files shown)


jbednar:
That's all very confusing, but I think it sounds ok.

ceball:
I would prefer to have a dev website that corresponds to the latest dev
package, and a release website that corresponds to the latest release, but
I'm not sure we agree about that.

I'm open to any better ideas...

jbednar:
I agree with that approach, except that there does often need to be a website update for the main, non-dev website on specific command.  I don't think that conflicts in any way, though.

ceball:
Ah, you meant the bit about updating the site is confusing. (I thought you were talking about the current package vs website problem and proposed solution.) Yes, I agree it's confusing; pyviz doesn't yet use our 'standard pyviz way' (that's what #81 is about), so I don't entirely remember how it works right now. But it looks like the website can be updated by any tag (on master).

And yes, in the 'pyviz standard way', there's a tag to update the website any time between releases (and also a tag to update the dev website any time between dev releases).

So about this PR...should we merge it and have pyviz.org saying to use `-c pyviz/label/dev` and people all getting the dev package for now, meaning what they get will match what's in the tutorials on pyviz.org? And then when #81 is done, go to the (roughly proposed) scheme above where pyviz.org corresponds to `-c pyviz` and (for instance) dev.pyviz.org corresponds to `-c pyviz/label/dev`?

I'm not really sure what's best. I like the idea that pyviz is a reliable way to get a working setup that matches some documentation, which I can confidently point people to, and isn't just another thing on top that is itself a problem to install and understand. But at the same time, I also like the idea that pyviz shows off all the latest stuff, plus I wouldn't want people to invest in older ways of doing things that will no longer be supported.

pyviz isn't 1.0 yet, so maybe it all doesn't matter too much.  (It could be 1.0 soon, though, right? I didn't see any new pyviz team package come out last week... :) )

jbednar:
That all sounds good, thanks.

ceball:
This change should appear on pyviz.org after travis finishes building the `website` tag.

jbednar:
Thanks!

Issue #100, opened 2018-09-17
Pip package


eleijonmarck:
Hello,

We would want to see the use case for intergrating pyviz into some of our tools. However we right all python.

Is there a way of installing everything without using `conda`

Preferably:
`pip install pyviz`

Or providing a makefile
`make install`

jbednar:
That's tricky -- even if you write pure Python, if you depend on anything that itself depends on things that are not Python, then you need to track binary dependencies, which Pip does not do very well.  That said, pyviz used to require conda specifically because Datashader required it, which in turn was because Numba required it, but now that numba ships via pip, Datashader also ships via pip, and it *may* be possible for pyviz to ship via pip.

@philippjfr, GeoViews would then be the key requirement.  I assume it would be difficult to get all the binary dependencies of GeoViews via pip?  Would it make sense for pyviz to be pip-installable *except* for GeoViews and its dependencies?

eleijonmarck:
@jbednar thank you for the consideration. We build a exploratory platform, however we have tried to combine `conda` with `pip` but were unsuccessful. Too many (weird) interactions too be able to maintain in production as well as setting it up for devs.

Perhaps a docker image would be better?

jbednar:
The conda team has been prioritizing improving interactions with pip, and there may already have been improvements there, if you use the latest conda releases.  The goal is to minimize those weird interactions, but I don't know if the specific ones you encountered were the ones that have been addressed so far.

You are certainly welcome to make a docker image and distribute it.  You should be able to set that up already now just by pip-installing the list of packages that pyviz includes, omitting any that don't have pip equivalents (which I think is just geoviews?).

ceball:
Some quick notes:
  * geoviews already has a package on pypi for pip, but it depends on cartopy, and whether you can pip install cartopy or not depends on how your system is set up (https://scitools.org.uk/cartopy/docs/v0.15/installing.html)
  * the geo stack is often problematic even if you use conda
  * a pip installable pyviz package will soon appear on pypi (as part of ongoing work), and I guess it will have optional extras (the core package will be fine to pip install)
  * if you're integrating pyviz into something else, do you need (or even want?) all of pyviz? Could you pip install only the underlying packages you need?
  * I imagine the kinds of conda/pip interactions being referred to are not improved in any released conda, *but* I think a preview of better conda/pip interaction might be an opt-in setting for one of the next conda releases

eleijonmarck:
[...]

PR #106, opened 2018-09-25
Panel: no moduled name panel

LuffyLuffy:
Hi there,
is that a version incompatible problem?
following is the result. I can't find answer of this problem.

$ pip install panel
Requirement already satisfied: panel in /anaconda3/lib/python3.6/site-packages (0.0.1)
Requirement already satisfied: param in /anaconda3/lib/python3.6/site-packages (from panel) (1.7.0)

$ python 
Python 3.6.6 |Anaconda custom (64-bit)| (default, Jun 28 2018, 11:07:29) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import panel
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'panel'


ceball:
Hi. Thanks for the report. pip packages for pyviz and all its dependencies are not yet available (see #100). Meanwhile, since you appear to be using Anaconda Distribution, you should be able to get everything you need via conda (see http://pyviz.org/installation.html). Please let us know if you have problems with that.

PR #111, opened 2018-09-28
Website fixups

ceball:
As noted in #81, I missed some newer changes; this PR fixes those.

I'm currently testing this PR, so might make more changes. I'll merge once I'm done.


ceball:
I can only think that I left around uncommitted/unpushed changes on another machine, because I'm sure I already made these replacements (except for the newer notebook)...

PR #114, opened 2018-10-02
Skip pip packages for now (see #100).

ceball:


ceball:
Seems like it should be uncontroversial (there are already no pip packages anyway), so I'll merge...

jbednar:
It's fine to skip them if there's some problem, but I thought we had agreed that we should now be able to have pyviz on pip.  I tried following the chain of issues and comments, but I can't see any fundamental remaining reason why we can't have pyviz on pip?

ceball:
I don't know about fundamentally, but practically speaking...

Current pyviz depends on unreleased packages not on pypi (e.g. bokeh 1.0 dev something).

Previous pyviz packages won't match the pyviz.org tutorial and will have broken pyviz --install-examples command. 

(what I wrote in https://github.com/pyviz/pyviz/issues/100#issuecomment-426395781)

jbednar:
Ok, sounds like we have to wait on Bokeh 1.0 non-dev to release for pip.  I'll try to remember that!

jlstevens:
Bokeh 1.0.4 is currently on [PyPI](https://pypi.org/project/bokeh/) and having a PyPI version of pyviz has both been requested by users #100 and is needed to get the `.version` file available for the conda-forge recipe.

ceball:
@jlstevens I tried (on linux) just now but pyviz requires geoviews and geoviews seemingly can't be pip installed from scratch - probably because cartopy is required by geoviews, and cartopy can't be pip installed from scratch (i.e. requires additional preparation first). If geoviews is required by pyviz and cartopy is required by geoviews, then I think additional instructions will be required for pip users.

Alternatively, what scope is there for making geoviews an optional extra? (Or cartopy an optional extra of geoviews?) I vaguely recall trying to start topics, e.g. "geo" (so pip `install pyviz[geo]` would be required to run the geo examples, kind of a thing)...but I guess I didn't get very far with that.

jbednar:
Our current plan for handling that is to [move WMTS map tiles support into HoloViews](https://github.com/ioam/holoviews/issues/3404), so that GeoViews becomes required only for http://pyviz.org/tutorial/07_Geographic_Data.html, letting people run the rest of the tutorials whether or not they have Cartopy and the rest of the geo stack installed. But that's not feasible for this release, so we'll just have to have additional instructions for pip users.

PR #118, opened 2018-10-10
Minor doc fixups

ceball:
Actually, one bit's probably not minor: as I came across the text, I decided to remove the list of individuals' names, since it duplicates information elsewhere (github) and is otherwise another thing to maintain (it's hard enough just to keep the list of pyviz-maintained packages up to date ;) ). I would instead be fine to revert that and update the list, but we'd need to keep it up to date. 

I also think we should also add a couple of bits: 
  * encourage contributions of all kinds from anyone
  * that pyviz and all the underlying tools are open source, free for any kind of use etc

I'll add that to this PR, if it seems reasonable?


jbednar (approved):
Adding those elements to the PR sounds good.  Removing the contributors seems fine.

doc/about.rst
```diff
@@ -7,8 +7,8 @@ including
 `HoloViews <http://holoviews.org>`_,
 `GeoViews <http://geo.holoviews.org>`_,
 `hvPlot <https://hvplot.pyviz.org>`_,
-`Datashader <https://github.com/bokeh/datashader>`_,
-`Param <https://github.com/ioam/param>`_, and
+`Datashader <https://datashader.org>`_,
```
ceball: That datashader link doesn't work for me...maybe it's down at the moment?

jbednar: Looks like it needs to be http, not https.

ceball:
@jbednar, which do you prefer: "import panel as pn" or "import panel as pp"?

```
apps/osm-1billion.py:import holoviews as hv, geoviews as gv, param, dask.dataframe as dd, panel as pn
pyviz/__init__.py:import panel as pp                                       # noqa (API import)
apps/nyc_taxi/main.py:import panel as pn
examples/tutorial/01_Workflow_Introduction.ipynb:    "import param, panel as pn\n",
examples/tutorial/12_Parameters_and_Widgets.ipynb:    "import param, panel as pp, numpy as np, holoviews as hv\n",
examples/tutorial/13_Deploying_Bokeh_Apps.ipynb:    "import holoviews as hv, geoviews as gv, param, dask.dataframe as dd, panel as pp\n",
examples/tutorial/A2_Dashboard_Workflow.ipynb:    "import panel as pp\n",
```

philippjfr:
We've settled on panel as pn.

jbednar:
Right; either was fine for me, but others voted for `pn`, so I've tried to fix any remaining `pp` that I've found.  Definitely needs a search and replace, though.

Issue #109, opened 2018-09-28
Can't run taxi player app on windows


ceball:
I noticed during #81 so can't investigate right now.

There seem to be multiple problems, but the first is some kind of decoding error. It's hard to say more, since I get only a redacted "ERROR" message from somewhere in the pyviz stack (bokeh?). My *guess* (it's really a guess) is somewhere in the pyviz stack opens the app's main.py file as text, and assumes the default encoding of text files is utf-8 - which is not true (at least, as of python 3.6 on windows). (The main.py file is fine to run with python itself rather than bokeh, since in python 3.6 python source code files are assumed to be utf-8 by default.)

Need to investigate further (e.g. does it happen with latest bokeh? does the app work on other platforms?).

marcbernot:
I had no such problem on window10  (fresh install with the latest miniconda python 3.7 then following the installation instructions from pyviz.org). We are talking about server_app.py right?

I already had encoding problems with the nyc_taxi_wide.parq file though. This happens by using pyarrow rather than fastparquet. Notice that when both are installed, pandas.read_parquet uses pyarrow by default.
`pd.read_parquet('nyc_taxi_wide.parq',engine='pyarrow')`-> encoding error
`df = pd.read_parquet('nyc_taxi_wide.parq',engine='fastparquet')`-> no error 
Writing a new parquet file with `df.to_parquet('new_nyc_taxi_wide.parq',engine='fastparquet')` yields a parquet file that can be read by both pyarrow and fastparquet.
I think you should replace nyc_taxi_wide.parq to get rid of this encoding error even if you recommend to work in a separate environment. I guess that some people just take the tutorials and try them in their already existing environment (this is what I did first).

ceball:
I was talking about the taxi player app. However, since I made this issue, that app has been changed to fix the first issue I encountered (although I think the fix is just hiding the underlying problem).

ceball:
I haven't revisited the app myself yet, so thanks for the info about the pyarrow/fastparquet issue. 

Your posts and investigations on pyviz have been very helpful, thanks.

jsignell:
That app no longer exists in the tutorial. So closing :)

PR #122, opened 2018-10-12
Updated instructions

jbednar:
Matches latest panel version.

Also restored apps apparently reverted in #81.


examples/tutorial/apps/taxi_player/main.py
```diff
@@ -74,14 +74,14 @@ def animate_update():
     slider.value = year
 
 def animate():
-    if button.label == '► Play':
-        button.label = '❚❚ Pause'
+    if button.label == '\u25B6 Play':
```
ceball, line 77: This doesn't seem like a fix, but a workaround for something broken elsewhere. Can you say what was broken? In #109 I found the app doesn't work on windows, and suspected/guessed the problem probably came from bokeh.

jbednar: See https://github.com/ioam/jupytercon2017-holoviews-tutorial/issues/5

ceball: Ok, seems like the same problem. I think this change should only be considered a temporary fix. While it doesn't matter much just for play/pause buttons in this app, people are going to struggle to read and write words using hex codes...

PR #124, opened 2018-10-19
First draft of updated 'about' page.

ceball:
Trying to help people understand what pyviz is and where support is available, in a brief way! Will likely need revising. Please feel free to edit.

Individual projects could use a part of this on their own 'about' pages, linking to pyviz.org's about page for more details. Otherwise I think we'll have too many pages to keep up to date...

### DIFF
--- a/doc/about.rst
+++ b/doc/about.rst
@@ -1,25 +1,89 @@
 About Us
 ========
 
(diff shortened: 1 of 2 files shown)


jbednar (approved):
Looks great to me, thanks!

PR #125, opened 2018-10-19
Fix testing of user install commands

ceball:
Added missing `--force`, because here we're using an already created directory. Currently, exactly following the docs and changing directory is an awkard thing to do with doit.


ceball:
Won't apply until next weekly job runs.

PR #128, opened 2018-10-30
Adding nyc_taxi topic from datashader

jsignell:
This is the first PR migrating datashader topics to pyviz topics. This PR uses the branch `pyct/support_test_files` to use small datafiles instead of large ones to run smoke tests.

Refer to `.travis.yml` in this PR to run website locally.

Topics index looks like: 

screen shot 2018-10-30 at 3 24 14 pm

Notebook succeeds using small dataset:

screen shot 2018-10-30 at 3 39 01 pm


doc/conf.py
```diff
@@ -30,7 +31,7 @@
     'PROJECT': project,
     'DESCRIPTION': description,
     'AUTHOR': authors,
-    # will work without this - for canonical (so can ignore when building locally or test deploying)
```
jbednar, line 33: Any idea what "for canonical" means?

examples/datasets.yml
```diff
@@ -6,6 +6,11 @@ data:
     files:
       - nyc_taxi_wide.parq
 
+  - url: https://s3.amazonaws.com/datashader-data/nyc_taxi.zip
+    title: 'NYC Taxi Data'
+    files:
+      - nyc_taxi.csv
+
```
jbednar, line 13: We'll probably want to eliminate the .csv version at some point, as it is slower and isn't meant to have any data not in nyc_taxi_wide.parq, but for now it's fine to have both.

BTW, can you please search for nyc_taxi_50k.parq in the tutorials/ ; I think at some point we tell people to use that manually if they don't have the big dataset, and they'll have to be pointed to the actual location of this file instead (add ".data_stubs/", which is nice and obvious that it's temporary).

examples/topics/index.ipynb
```diff
@@ -0,0 +1,48 @@
+{
```
jbednar, line 1: I don't think there's any reason to commit the export/ directory, is there?  It's not used on the website; those images are just for me to supply to people who want screenshots.

examples/topics/index.ipynb
```diff
@@ -0,0 +1,48 @@
...
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.6.6"
```
jbednar: Please use strip_kernel (from Jean-Luc) to clear metadata before committing new notebooks.

jbednar (approved):
Looks great, thanks!

doc/conf.py
jsignell: nope :(

examples/datasets.yml
jsignell: Oh I see that. I am adding it back in.

examples/topics/index.ipynb
jsignell: Ok that makes sense.  But why is it generated there at all? Shouldn't it be only in temporary files?

examples/topics/index.ipynb
jsignell: whoops. Yep doing that now.

examples/topics/index.ipynb
jbednar: Those aren't temporary files, they are exported versions -- they are meant to be left at the end, for people (typically me) to use outside of the notebook context.  We can disable them altogether, but for a time I was having people request screenshots all the time, so I had them generated as a side effect.

doc/conf.py
ceball: The canonical url, for search engines.

Same text appears in multiple projects. Maybe a bit late now, but the original source of the note is https://github.com/pyviz/nbsite/blob/master/nbsite/tmplate2/conf.py#L39 (template used to start a new project).

jbednar: Ok, thanks.  @jsignell, please change the comment to "WEBSITE_SERVER is optional for tests and local builds, but allows defining a canonical URL for search engines"

jsignell:
Can't figure out how to restart appveyor builds now that I've released pyct, but if Travis passes I'll push a little commit.

jsignell:
Ok I think this will be ready when it goes green.

jsignell:
hmmmmm looks like appveyor timed out.

jbednar:
Hmm. It seems to have been killed after 1 hour, yet the logs only show conda commands. Surely setting up the conda envt didn't take that whole hour?  I'll try committing some tiny change to this branch to try again; I don't know any other way to force an appveyor build either.

jsignell:
Thanks @jbednar :) I think this is good to go. Should I merge?

PR #130, opened 2018-11-02
Updated homepage with new logos and categories

jbednar:
The previous pyviz.org homepage had a partial listing of some of the supported libraries, without any way for the user to see that they were each supported in very different ways.  This PR adds many more logos and sorts them into meaningful categories:

image

image


philippjfr:
Looks fabulous to me, much better than putting them all together. My only comment/question is whether we should consider moving param to the top, I can see arguments either way though.

jbednar:
I'd add logos for geopandas, colorcet, vega, and vega-lite, but none of those have proper logos yet.

philippjfr:
It doesn't make an exceptionally great logo but geopandas does have a logo: https://pbs.twimg.com/profile_images/507620807242039297/UhLrFw_Q_400x400.png

jbednar:
> My only comment/question is whether we should consider moving param to the top, I can see arguments either way though.

Yes, I've tried it both ways, and Jean-Luc and I have already discussed that.  It definitely doesn't fit where it is, and it's a bit buried there, but I'm not sure we want to present it as a "core high-level library". Is it something we want people to *start* with, or *end up*  with?  At the moment I've leaned towards the latter, which is why it is down there, but I think it's worth discussing and maybe voting.

jsignell:
This looks great! I was trying to write something similar - but this is much more comprehensive. My only comment is it seems like there should be a more visual way of presenting the relationships between the libraries, maybe as building blocks or something? That can obviously be added later if someone makes it.

jbednar:
I've moved Param to the first item in Other, so that it stands out better, and explained it slightly more.  I also added a GeoPandas logo as suggested, created a Vega logo out of nothing, and updated the FAQ and Roadmap to reflect Panel and other recent improvements.

> There should be a more visual way of presenting the relationships between the libraries, maybe as building blocks or something.  

Yes, but it's tricky.  The usual such diagrams are just based on layers, where one thing builds on another, but here the packages have complex relationships to each other that don't quite fit that organization.  We probably do need such a figure for giving talks, and I'll probably make one the next time I give a talk, but at least the various packages now (a) have logos of the right sort and (b) are in meaningful groups, which is a starting point.

PR #131, opened 2018-11-02
First use of nbsite --skip, automatically touching .nojekyll file, cleaning arg added

jsignell:
Won't be mergeable until after https://github.com/pyviz/nbsite/pull/93 lands and a version is bumped,
but first use of `nbsite generate-rst --skip` and not needing to touch .nojekyll anymore. Also cleaning up dist files be default.


jsignell:
I think this is finally mergeable :)

PR #137, opened 2018-11-29
Switch to using full dataset for website build

jsignell:
Locally the fetch data took about 250sec and the website build took another 250sec. So this should be fine while we iron out best practices for using an evaluated notebook branch (see earthml for that method)


jsignell:
@jbednar @jlstevens should I merge this and tag a release?

PR #141, opened 2018-12-15
Avoid full rerender in NYC taxi example

philippjfr:
@jbednar I'm not sure how we can make the difference here clearer, but triggering a full rerender on each parameter change is definitely not desirable. I kind of regret not pushing back on the idea that an unannotated method implicitly depends on all parameters because of this.


jbednar:
As far as I can tell, this change needs to be reverted -- it's no longer updating the plot when the widgets are used, because it doesn't call `make_view` in that case any more. I believe it was already correct, though (as explicitly acknowledged in the text just after this example) not very efficient.

PR #142, opened 2018-12-17
Use panel interact for first NYC Taxi example

philippjfr:
I was editing an outdated version of the notebook so let me know if you want to restore some of the newer text @jbednar.


examples/tutorial/01_Workflow_Introduction.ipynb
```diff
@@ -699,41 +699,32 @@
...
-    "pn.Row(explorer.param, explorer.make_view()).servable()\n"
+    "pn.Row(pn.Column('# NYC Taxi Explorer', explorer[0]), explorer[1]).servable()"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Note that in this simple app, the method `make_view` is called whenever *any* of the parameters change (alpha, colormap, or location), but you can get a more responsive interface if you take the time to declare which computations depend on which parameters (see the [Deploying Bokeh Apps tutorial](./13_Deploying_Bokeh_Apps.ipynb)).\n",
+    "Note that in this simple app, the method `make_view` is called whenever *any* of the parameters change (alpha, colormap, or location) triggering a full rerender, but you can get a more responsive interface if you take the time to clare which computations depend on which parameters (see the [Deploying Bokeh Apps tutorial](./13_Deploying_Bokeh_Apps.ipynb)).\n",
```
jbednar:
```suggestion
    "Note that in this simple app, the method `make_view` is called whenever *any* of the parameters change (alpha, colormap, or location) triggering a full rerender, but you can get a more responsive interface if you take the time to declare which computations depend on which parameters (see the [Deploying Bokeh Apps tutorial](./13_Deploying_Bokeh_Apps.ipynb)).  Either way, the app should work the same here in the notebook (if you have a live Python process) or as a standalone server by calling `panel serve` with either the name of a Python file with the above code or simply the name of this notebook (where it will serve any objects marked `.servable()`).)"\n",
```

examples/tutorial/01_Workflow_Introduction.ipynb
```diff
@@ -699,41 +699,32 @@
    "source": [
     "As you can see, you can specify geo plots easily with GeoViews, and if your HoloViews objects are too big to visualize in a browser directly, you can add `datashade()` to render them into images dynamically on zooming, etc.\n",
     "\n",
-    "You can also easily add widgets to control filtering, selection, and other options interactively, either here in the notebook, or in a standalone server by marking the servable objects with `.servable()` then running the .ipynb file through Bokeh Server or extracting the code to a separate .py file and doing the same thing:"
+    "You can also easily add widgets to control filtering, selection, and other options interactively, either here in the notebook or by putting the same code (plus `panel.servable();`) in a separate file and running it as a standalone server:"
```
jbednar:
```suggestion
    "You can also easily add widgets to control filtering, selection, and other options interactively to make apps or dashboards:"
```
(rest of the text moved to comment below.)

examples/tutorial/01_Workflow_Introduction.ipynb
jbednar: (Not sure if anyone can simplify that mouthful of text!

examples/tutorial/01_Workflow_Introduction.ipynb
```diff
@@ -699,41 +699,32 @@
...
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
-    "import param, panel as pn\n",
-    "from colorcet import palette\n",
+    "def make_view(location='dropoff', cmap=palette['fire'], alpha=1):\n",
+    "    pts = hv.Points(taxi, [location+'_x', location+'_y'])\n",
+    "    trips = datashade(pts, cmap=cmap)\n",
+    "    return tiles.options(alpha=alpha) * trips\n",
```
jbednar: Doesn't it still need to import panel and colorcet?

examples/tutorial/01_Workflow_Introduction.ipynb
```diff
@@ -699,41 +699,32 @@
...
   },
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "slideshow": {
-     "slide_type": "slide"
-    }
-   },
+   "metadata": {},
```
jbednar: Please restore the slideshow metadata in that cell.

jbednar (changes requested):
Much simpler, thanks!

examples/tutorial/01_Workflow_Introduction.ipynb
jbednar: Never mind -- I'll edit it locally and push to this branch.

philippjfr:
Thanks! Otherwise I'll get to it tonight.

jbednar:
Ok, I'm happy with it.  Much simpler as a first example!

jbednar:
BTW, I think you mentioned that it should point to something other than 13_Deploying_Bokeh_Apps.ipynb , and if so, please update that link.

jbednar (approved):

philippjfr:
I was just confused about that, I thought that was a link to holoviews docs.

jbednar:
Ok.  I'm pretty sure what it does link to needs updating for current approaches, of course.  I've updated to panel>=0.3 in the environment; hopefully that will make the tests pass.

jbednar:
I give up on the environment; not sure if there is any recent combination that will work. May need a dev release of something?

philippjfr:
Looks like I should tag a new hvplot dev release.

setup.cfg
```diff
@@ -105,30 +101,27 @@ universal = 1
...
+    streamz = 0.5.0
     # for jupyter notebook and bokeh server
     tornado = 4.5.3
     # in case of unknown future or past compatibility issues
     python = 3.6.*
     numpy = 1.14.5
-    pandas = 0.23.4
-    xarray = 0.10.3
+    pandas = 0.24.0
+    xarray = 0.11.0
```
jsignell: It'd be nice to have 0.11.2 if it's available - adds some nice features for netcdf - but it's not on defaults yet, so 
could we just have a floor pin?

```suggestion
    xarray >= 0.11.0
```

philippjfr: I'm actually not totally clear what this section is about, I thought this was only for running the tests where I'd prefer a fixed pin.

ceball: The pins section is used by pyctdev when `--pin-deps` is passed. 

`--pin-deps` was being used for running the tests and for generating packages (i.e. to generate the highly pinned pyviz packages).

Using it for the tests was removed in #152. I don't know if you are planning to continue to release highly pinned packages or not.

.travis.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       before_script:
         - doit env_create --name=pyviz --python=$PYENV_VERSION
         - source activate pyviz
-        - travis_wait 20 doit develop_install $CHANS_DEV -o recommended
+        - travis_wait 20 doit develop_install $CHANS_DEV -o tests -o examples -o graphs -o geo -o indirect
```
ceball: This was previously `--all-extras`, but that was removed in #152 and replaced with just `-o recommended`. I can't quite follow what the intention is.

philippjfr: recommended isn't including fastparquet, which is listed under `indirect` dependencies. The correct solution is to move fastparquet in another section if I was guessing.

ceball: Short version:

Unless you're trying to change how things work in some way (which I'd be fine with, but you'd need to give me some background for me to help out with that), I *think* some of #152 should be reverted:
  1. Restore the pinning and use of 'all extras': https://github.com/pyviz/pyviz/pull/152/files#diff-354f30a63fb0907d4ad57269548329e3L49
  2. Restore installation of pyctdev from `-c pyviz/label/dev`

I.e. the important bit to keep is the minimum version pin of pyctdev to avoid a very old version of pyctdev being installed.


Long version!

I *think* #152  might have come from a misunderstanding, which comes from an unfortunate conjunction of complications:

  1. Originally pyctdev did not pin conda build. However, that didn't work out well because of variability in conda-build behavior over time. So pyctdev pinned a specific conda build. But...that version of conda build has no python 3.7 build. 
  2. When you install miniconda these days you get python 3.7. Until recently, conda install pyctdev in a fresh miniconda install would go for a recent pyctdev and downgrade python 3.7 to 3.6. That was conda <=4.5.11 I think. conda 4.5.12 changes that to instead installing an old pyctdev (presumably from before conda build was pinned by pyctdev) and sticking with python 3.7.

The above only affects the root/base environment (i.e. you can still make and test python 3.7 packages with pyctdev).

ceball: And yes, pyctdev should be updated to use a newer conda-build. I am working on that. Unfortunately I am working on that at the same time as a new version of pyctdev - the purpose of which is to stop using some problematic/outdated features of conda build.

.travis.yml
```diff
@@ -42,11 +42,11 @@ jobs:
         - pip install pyctdev && doit miniconda_install && pip uninstall -y doit pyctdev
         - export PATH="$HOME/miniconda/bin:$PATH" && hash -r
         - conda config --set always_yes True
-        - conda install -c pyviz "pyctdev>=0.5" pyctdev && doit ecosystem_setup
+        - conda install -c pyviz/label/dev "pyctdev>=0.5" pyctdev && doit ecosystem_setup
       before_script:
         - doit env_create --name=pyviz --python=$PYENV_VERSION
         - source activate pyviz
-        - travis_wait 20 doit develop_install $CHANS_DEV -o tests -o examples -o graphs -o geo -o indirect
+        - travis_wait 20 doit develop_install $CHANS_DEV -o recommended --all-extras
```
ceball: I think you should remove `-o recommended` as it isn't doing anything. I think pyctdev should be giving an error for this, as there's no `recommended` set of extras defined in setup.cfg.

ceball: I filed a bug report: https://github.com/pyviz/pyct/issues/73

philippjfr: Thanks, so I should define no ``-o`` at all since ``--all-extras`` is already doing it?

setup.cfg
```diff
@@ -105,30 +101,27 @@ universal = 1
...
+    geoviews = 1.6.2
+    datashader = 0.6.9
+    panel = 0.4.0
+    param = 1.8.2
     ###
     # avoid tab-completion issue in ipython=6
     ipython = 5.*
     # for datashader compatibility
-    dask = 0.18.2
+    dask = 1.1.0
```
ceball, line 112: Is this available on defaults? If not, you'll either need to use e.g. conda-forge (in CI and in instructions for users), or copy dask onto the pyviz channel...

ceball: (By defaults I mean the anaconda main channel, sorry.)

philippjfr: I'll probably relax the pin instead.

.travis.yml
philippjfr: Getting an error with `doit develop_install $CHANS_DEV --all-extras`

philippjfr: The main thing I'm confused about here is why ``-o examples`` doesn't work, in the tox.ini that defines all the correct dependencies. Is it getting confused about the ``examples`` entry in `extras_require`?

ceball: > Thanks, so I should define no -o at all since --all-extras is already doing it?

Right

> Getting an error with doit develop_install $CHANS_DEV --all-extras

The error being the following? (I can't read as fast as you can code 😂) 

```
PackagesNotFoundError: The following packages are not available from current channels:
  - fastparquet[version='>=0.2.1']
Current channels:
  - https://conda.anaconda.org/pyviz/label/dev/linux-64
  - https://conda.anaconda.org/pyviz/label/dev/noarch
  - https://repo.anaconda.com/pkgs/main/linux-64
  - https://repo.anaconda.com/pkgs/main/noarch
  - https://repo.anaconda.com/pkgs/free/linux-64
  - https://repo.anaconda.com/pkgs/free/noarch
  - https://repo.anaconda.com/pkgs/r/linux-64
  - https://repo.anaconda.com/pkgs/r/noarch
```

Or something else?

philippjfr: Haha, yes, I just didn't look closely enough thanks, that makes sense now. If you can explain to me why ``-o examples`` didn't work (see https://travis-ci.org/pyviz/pyviz/builds/486033035), i.e. didn't install the 'indirect' dependencies listed in the setup.cfg, I think I'll finally understand how some of this stuff works.

setup.cfg
philippjfr: >If not, you'll either need to use e.g. conda-forge (in CI and in instructions for users), or copy dask onto the pyviz channel...

Actually we also want latest fastparquet, I'll add conda-forge to the CI now and then hope that the conda folks update fastparquet and dask along everything else this week.

setup.cfg
philippjfr: Thanks Chris, it's all starting to fit together for me. I'll pin 0.11.2 now that I've added conda-forge.

setup.cfg
jsignell: Please don't depend on c-f unless its needed.

jsignell: Are we depending on things that are in the most recent versions?

philippjfr: It is until fastparquet and dask are updated in defaults unless we want to force user to use old pandas.

philippjfr: Alternatively we could also stop pinning pyviz as strictly as we have been.

jsignell: I would vote for not pinning as tightly. Or pinning between a few versions

.travis.yml
philippjfr: Nevermind I see that the targets in tox.ini have nothing to do with it.

jsignell: yeah sorry I misled you :|

setup.cfg
jbednar: The priority right now is to get the pyviz package released, and it seems fine to depend on c-f for the CI builds and installations (for now) until all the dependencies are on defaults.  The channels don't go into the package definitions, and once things are available we can update the CI not to use c-f.

jbednar: For pinning it's fine to pin a version range, but all such versions should be ones we've actually tested, which means not allowing *future* versions (at least not new major or minor versions).

jbednar:
Should we consider these pins and dependencies in setup.cfg?
- pandas >= 0.23.4,<=0.24.1 ?  (allowing Pandas one bugfix release that we trust to be the same?)
- ipython >= 5, <=7.1.1  (for tab completion)
- cartopy    ?   (didn't we have problems with a recent cartopy release, that mean we should pin it?)
- intake
- intake-xarray

intake-xarray has a zarr dependency only available from conda-forge. The current pyviz examples don't use intake, but we have others floating around that I want to be able to depend easily on intake.  I'm not *certain* this is appropriate, but it seems like something worth doing?

examples/tutorial/01_Workflow_Introduction.ipynb
```diff
@@ -684,12 +684,15 @@
...
-    "trips = datashade(pts, cmap=fire, **args)\n",
+    "colormaps = {n: palette[n] for n in ['fire','bgy','bgyw','bmy','gray','kbc']}\n",
+    "\n",
+    "def view(location='dropoff', cmap=colormaps['fire'], alpha=1):\n",
+    "    pts = hv.Points(taxi, [location+'_x', location+'_y'])\n",
+    "    trips = datashade(pts, cmap=cmap)\n",
+    "    return tiles.options(alpha=alpha) * trips\n",
     "\n",
-    "tiles * trips"
+    "view()"
```
jbednar, line 704: See https://github.com/pyviz/panel/issues/227 ; may want to rename "view".

jbednar:
@Philipp, I'm now fine with this version in terms of changes to the tutorial notebook; I made sure it works as a slideshow again (which required some minor fixups) and updated a bit of the text.  So now it can be merged as soon as the envt is ok, and I think that's all that's needed for the release.

One more thing to consider changing: replace "-c pyviz/label/dev" in doc/installation.rst with "-c pyviz".  I haven't changed it as the dev channel should still work, and I'm not sure if we prefer to have pyviz floating up with dev releases as we make them for now.  Seems ok either way.

philippjfr:
>cartopy ? (didn't we have problems with a recent cartopy release, that mean we should pin it?)

Not cartopy itself though it was just GDAL issues, I'd also expect that pinning to happen in geoviews.

setup.cfg
philippjfr: I'll leave the list of strict pins in the setup.cfg for now, so we can potentially a heavily pinned ``pyviz-tutorial`` package for running live tutorials.

philippjfr:
This appears to be mostly working but one cell keeps timing out:

```python
import geopandas as gpd
nworld = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))
nworld.head()
```

This should be very quick.

jsignell:
Does that depend on the data already being cached?

philippjfr:
>Does that depend on the data already being cached?

No, it should download it, maybe that's really slow? I tried increasing the cell timeout with this in the setup.cfg:

```
[pytest]
nbsmoke_cell_timeout = 600
```

Don't think that worked though.

philippjfr:
I've skipped the notebook now. So we could merge and tag a dev release. conda-forge is not used for real releases though so we either have to add it for those as well or wait until at least fastparquet 0.2.1 has made it into defaults (xarray and dask pinning were relaxed).

examples/tutorial/01_Workflow_Introduction.ipynb
jbednar: Should be fine; "view" here is a verb, and that's the proposed usage.

PR #152, opened 2019-01-28
Updated pinning of pyctdev

jlstevens:
Hopefully this will fix the CI issues.


jlstevens:
My changes seem to have gotten the tests running even though many of the notebooks are failing (in the smoke tests I believe).

One other thing I noted in the log is that Python is being downgraded:

```
 python:          3.7.1-h0371630_7             --> 3.6.8-h0371630_0      
```

@jbednar seems to remember a similar issue with failing smoke tests due to those tests being run in an inappropriate environment. @jsignell do you have any suggestions how to proceed?

Skimming through the failures I think I saw one that might be genuine. Checking that now.

jbednar:
Looks like an improvement, anyway.

Issue #144, opened 2018-10-22
Panel user guide shows import panel as pn; pyviz shows import panel as pp


ahuang11:
Which one is preferred?

jlstevens:
Addressed in the PR mentioned above. @ahuang11 thanks for reporting this!

PR #153, opened 2019-01-28
Importing panel as pn and not pp

jlstevens:
Addresses #144


jlstevens:
Very simple change. Merging.

PR #157, opened 2019-02-07
Trying out a new style of readme

jsignell:
Loosely based on the pandas readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,14 +6,10 @@
 
 |    |    |
 | --- | --- |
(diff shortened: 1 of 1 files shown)


jbednar:
Looks good!

image

But I'd vote for laying out the badges in each  labeled category left to right in a single row, rather than adding multiple rows.  

Also, "Latest dev release" is a bit misleading, as I think that shows the latest tag?  Maybe add a badge for pyviz/label/dev's current version, alongside the tag one, which would make it more accurate?

jsignell:
image

jsignell:
Any other things that would be helpful? The vision is that a person landing in a particular project will see these badges and we will be able to get a broad overview of the same information on the status-dashboard.

jbednar:
Looks great, thanks so much!

PR #162, opened 2019-03-08
Expanded background section with shortcuts/dead-ends discussion

jbednar:
Should also eventually move tutorial 01 into the background, as it is more bg info than strictly a tutorial, but that will take some restructuring...


doc/background.rst
```diff
@@ -32,20 +32,24 @@ exploration. Bokeh and Matplotlib both also have limitations on how much
...
 
-To address these issues, we have developed a set of open-source Python
+
+The PyViz ecosystem
+-------------------
+
+To address all the above issues, we have developed a set of open-source Python
 packages to streamline the process of working with small and large datasets
-(from a few points to billions) in a web browser, whether doing
+(from a few points to billions or more) in a web browser, whether doing
```
jsignell: Sounds a little like we only deal with points.

doc/background.rst
```diff
@@ -32,20 +32,24 @@ exploration. Bokeh and Matplotlib both also have limitations on how much
...
 -  `Panel <http://panel.pyviz.org>`__: Assembling objects from
    many different libraries into a layout or app, whether in a Jupyter
-   notebook or in a standalone serveable dashboard
+   notebook or in a standalone servable dashboard
 -  `Bokeh <http://bokeh.pydata.org>`__: Interactive plotting in web
    browsers, running JavaScript but controlled by Python
 -  `hvPlot <http://hvplot.pyviz.org>`__: Quickly return interactive
-   HoloViews or GeoViews objects from your Pandas, Xarray, or other
-   data structures
+   Bokeh-based HoloViews or GeoViews objects from your Pandas, Xarray,
```
jsignell:
```suggestion
   Bokeh-based HoloViews or GeoViews objects from Pandas, Xarray,
```

doc/background.rst
```diff
@@ -57,25 +61,100 @@ full-featured dashboards. The libraries in this ecosystem include:
...
 
-Here Datashader is used to make large datasets practical by rendering
-images outside the browser, either directly for a programmer or via a
-convenient high-level interface using HoloViews (or hvPlot), and the
-results can be embedded in interactive Bokeh plots if desired, either
-as a static HTML plot, in a Jupyter notebook, or as a standaline
-application.
+Here both users are using Datashader, but in very different ways.  The
+data analyst could use the high-level interface from HoloViews or
+hvPlot, which can use Datashader when needed for large datasets,
```
jsignell:
```suggestion
hvPlot, implicitly using Datashader when needed for large datasets,
```

doc/background.rst
```diff
@@ -57,25 +61,100 @@ full-featured dashboards. The libraries in this ecosystem include:
...
+then be used in web browsers directly as static HTML, in Jupyter
+Notebooks for interactive sessions, or as a standalone separate
+dashboard or application. In each case the analyst pulls in some part
+of the ecosystem, as needed.
+
+Meanwhile, a dedicated software developer might use some of these same
+tools, perhaps by calling Datashader directly to generate bare images,
+and then optionally embedding those images into an interactive Bokeh
+plot (or perhaps directly into a report or a web page). This example
+covers only a few of the available tools , and each user can apply any
```
jsignell:
```suggestion
covers only a few of the available tools, and each user can apply any
```

doc/background.rst
```diff
@@ -57,25 +61,100 @@ full-featured dashboards. The libraries in this ecosystem include:
...
+precisely control how it works. For PyViz, Bokeh is a low-level tool,
+allowing any plot or app to be built up from basic primitives. A
+high-level tool like hvPlot or like Panel's interact function uses
+much less code and a much simpler interface to make a powerful plot or
+dashboard, but what do you do if you then need to make some small
+changes in the result?
+
+Typical high-level tools will just be dead ends at this point, forcing
+you to start over if what they provide isn't what you needed. PyViz
+high-level tools are instead very systematically designed as layers on
```
jsignell:
```suggestion
high-level tools are instead systematically designed as layers on
```

doc/background.rst
```diff
@@ -57,25 +61,100 @@ full-featured dashboards. The libraries in this ecosystem include:
...
 
-Behind the scenes, these tools rely on a wide range of other open-source
-libraries for their implementation, including:
+.. image:: assets/shortcuts.png
+    :height: 300px
+
+A low-level tool is highly configurable, with an expressive but
+relatively verbose command language that makes it possible to
+precisely control how it works. For PyViz, Bokeh is a low-level tool,
+allowing any plot or app to be built up from basic primitives. A
```
jsignell: Mention that using js directly would be even lower level.

doc/background.rst
```diff
@@ -57,25 +61,100 @@ full-featured dashboards. The libraries in this ecosystem include:
...
+Panel's `interact function <https://panel.pyviz.org/user_guide/Introduction.html>`_
+provides a clear example of this approach. With one line of Panel code
+you can get a fully functional widget-based app.  But if it's not
+precisely what you want, you can then inspect what's returned, unpack
+it, rearrange and add or delete components, then use the result
+instead.  Similarly, hvPlot provides a one-line way to return complex
+HoloViews objects, which can then be inspected, pulled apart,
+reconfigured, and recomposed if/as needed. And then these HoloViews
+objects, in turn, can be used to make a Bokeh figure that again can be
+examined, modified, and used in other contexts (if desired!).
```
jsignell, line 150: Is this true? I think you'd have to add a hook to the original function.

doc/background.rst
```diff
@@ -57,25 +61,100 @@ full-featured dashboards. The libraries in this ecosystem include:
...
 .. image:: assets/ds_hv_bokeh.png
     :height: 220px
 
-Here Datashader is used to make large datasets practical by rendering
-images outside the browser, either directly for a programmer or via a
-convenient high-level interface using HoloViews (or hvPlot), and the
-results can be embedded in interactive Bokeh plots if desired, either
-as a static HTML plot, in a Jupyter notebook, or as a standaline
-application.
+Here both users are using Datashader, but in very different ways.  The
```
jsignell, line 82: I think if you weren't already familiar with these tools this section would be hard to follow. Maybe try to emphasize more the differences between the approaches. Also the image above doesn't show hvplot.

jsignell (commented):
I think the added content generally makes sense. I have some small comments.

jsignell:
@jbednar would you like me to fix up this PR, and apply the changes I suggested?

doc/background.rst
jbednar: Any better, where "datapoints" are meant to be more general than "points"?
```suggestion
(from a few datapoints to billions or more) in a web browser, whether doing
```

jbednar:
Yes, please do! Thanks.

doc/background.rst
jsignell:
```suggestion
allowing any plot or app to be built up from basic primitives. An
even lower-level approach would be to write javascript directly. A
```

PR #169, opened 2019-04-04
:memo: Switch to protocol-relative URLs

rs2:
The docs should use `https://` where possible, but even better, default to the protocol of the current page.


philippjfr:
Thanks for doing this @rs2. I'm happy for this to be merged.

philippjfr:
Not sure what's up with the tests.

jbednar (commented):
What are the advantages and disadvantages of this approach?  E.g. what happens when these files are viewed in a local clone of the repository -- won't they then (incorrectly) access the links as file:// links?

jsignell:
I tested it out locally and it doesn't seem to work when previewing the file from an editor. I think being able to see the icons from an editor is a nice feature for the README, but maybe less important for the others.

jbednar:
But it's not just the icons, right?  Won't *every* external link (to content outside this particular site) be broken locally, e.g. for everyone using our Jupyter notebooks?  That doesn't seem like an improvement.

jsignell:
Yes that is true. Thanks for the suggestion though @rs2! It might be something that would be better incorporated into nbsite. And we do need to do a pass and make sure we are linking to the https versions of all available sites.

jbednar:
Oh, that's a good point.  We could make the built website use protocol-relative URLs without changing the underlying source, essentially doing what this PR does but dynamically when building the site.  That would address my objections...

PR #184, opened 2019-06-13
Fixing channel url

jsignell:
The short channel name conflicts with the local package name.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: generic
 env:
   global:
     - CHANS_DEV="-c conda-forge -c defaults -c pyviz/label/dev"  # this order matters because we're using strict channel priority
-    - CHANS_REL="-c conda-forge -c defaults -c pyviz"            # this order matters because we're using strict channel priority
+    - CHANS_REL="-c conda-forge -c defaults -c pyviz/label/main"            # this order matters because we're using strict channel priority
     - LABELS_DEV="--label dev"
     - LABELS_REL="--label dev --label main"
     - PYENV_VERSION=3.7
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,8 +2,8 @@ language: generic
 
 env:
   global:
-    - CHANS_DEV="-c conda-forge -c defaults -c pyviz/label/dev"  # this order matters because we're using strict channel priority
-    - CHANS_REL="-c conda-forge -c defaults -c pyviz"            # this order matters because we're using strict channel priority
+    - CHANS_DEV="-c conda-forge -c defaults -c pyviz/label/dev"   # this order matters because we're using strict channel priority
+    - CHANS_REL="-c conda-forge -c defaults -c https://conda.anaconda.org/pyviz"  # this order matters because we're using strict channel priority
```
jbednar, line 6: Why would we not want the pyviz channel to be in front, to give a quick and accurate answer for the things that are on it?

Also, why would a channel named "pyviz" have any conflict with a package named "pyviz"? Is that a bug in conda?

jsignell: > Why would we not want the pyviz channel to be in front

I am doing strict channel priority in these jobs, so if there were any package that was added to the pyviz channel (like phantomjs, or rise) and then that package got out of date, we'd still get that one. Since we are pinning all of the pyviz packages very closely, we will get the ones we want even with lower channel priority on the pyviz channel. 

> Also, why would a channel named "pyviz" have any conflict with a package named "pyviz"? Is that a bug in conda?

I'm not sure if these was an issue with pyctdev or with conda, but it only comes up when installing from local builds. This change was just a reversion to how the travis was when the last release passed CI. I probably changed it in an attempt to simplify without really knowing why it was using the full url.

PR #204, opened 2019-07-08
DOC: tweak the wording around Matplotlib to be more accurate

tacaswell:
The inline backend is static, however nbagg (which ships with
Matplotlib, can be activated via `%matplotlib notebook` and works in
classic notebook) and ipypml (which is an independent package, can be
activated via `%matplotlib widget` and works in both classic notebook
and jupyterlab) provide interactive backends for the notebook.


jsignell (approved):
Thanks for the PR! We have a note to update the matplotlib parts to try to show the best possible comparisons.

tacaswell:
Thanks!

PR #206, opened 2019-07-09
Setting up binder

jsignell:
See https://mybinder.org/v2/gh/pyviz/holoviz/jsignell/binder_data?filepath=examples
closes #206


jsignell:
Note that the kernel dies pretty regularly when large data are persisted, but maybe we could spin up our own instance that doesn't do that if we need.

jbednar:
Done in another PR. Thanks!

PR #217, opened 2019-08-29
Fixing typo in environment.yml and adding to tests

### LINKED ISSUE #215: xarray version incorrect
I'm trying to replicate (manually) the environment needed for the tutorials and came across this issue.

In the `setup.cfg` and the `environment.yml` file in the repository `xarray=0.12.1A` is specified but doesn't exist either in the pyviz, anaconda or conda-forge channels.

jsignell:
Closes #215

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,7 +66,11 @@ jobs:
         - source activate holoviz-env
         - conda install -c pyviz -c defaults nbsmoke pytest flake8
         - doit env_capture
-
+      script:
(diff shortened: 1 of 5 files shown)


jsignell:
Note to self: need to think more about whether we want to test using the env. There are no holoviz commands if you take that route. So we'd need a different way of fetching the data (could probably use pyct directly)

jsignell:
Note I added a line in the setup notebook that will do the data fetch for you. So it should now be feasible to just download the examples dir and get cracking.

PR #220, opened 2019-09-04
Use gzip rather than snappy

jsignell:
This makes it so that all dependencies are pip-installable. Plus the file is smaller. I used a different name so as not to break the existing tutorial.


jsignell:
@jbednar I can't imagine a downside of this. Should I merge?

jbednar:
> I can't imagine a downside of this. 

Oh, there is always a downside. :-)  See: 

https://github.com/pyviz/datashader/issues/129#issuecomment-299605103

At least on that benchmark (census data) at that time, gzip took twice as long overall as snappy did.  That doesn't mean we shouldn't do it to make running the examples smoother, but if similar results are found here then that's definitely a downside.

jsignell:
As soon as I wrote that I knew there would be something...

jsignell:
However for a relatively small file, maybe having a pip installable tutorial is worth the slow down in decompression?

jbednar:
Can you test the performance difference?  It might be fine for this case.

jsignell:
Seems that gzip is marginally faster:

Screen Shot 2019-09-04 at 4 51 01 PM

jbednar:
Earlier we found some very surprising effects for Dask dataframes where the compression type affected not just the loading times but the subsequent aggregation times.   I'm not sure we ever explained it, but please add an aggregation step like in the original benchmark, not just .compute(); that way it would be a more realistic model of what people actually do in our examples.

jsignell:
It definitely takes longer when not using snappy, but not sure if the aggregation is the slow step in this case compared to the rendering. 

``` 
# dask
nosnappy.parq    Aggregate1:003.31 (002.89+000.43)  Aggregate2:000.02  In:00094104374  Out:00000119374  Total:008.63
y_snappy.parq    Aggregate1:002.78 (002.39+000.39)  Aggregate2:000.01  In:00163701025  Out:00000119374  Total:007.78

# pandas
nosnappy.parq    Aggregate1:008.24 (007.79+000.45)  Aggregate2:000.03  In:00094104374  Out:00000119374  Total:021.90
y_snappy.parq    Aggregate1:007.36 (006.91+000.46)  Aggregate2:000.03  In:00163701025  Out:00000119374  Total:021.91
```

jbednar:
The important thing to focus on here is the Aggregate2 timings, because that's what affects interactive response times for a user who is panning and zooming, and thus determine how easily someone can explore a dataset of a given size.  Here, with dask the reported Aggregate2 times are twice as fast with snappy, which if real would be a big deal and a strong reason to continue using snappy.  But it's not really possible to tell because of how fast the aggregation completes in this case; it could be just a rounding issue with times that are actually quite close.  I can't remember how the benchmark worked; is it easy to increase the task size so that Aggregation2 takes long enough to avoid the rounding errors?

jbednar:
Pushed separately to master. Thanks!

PR #221, opened 2019-09-10
Adding flow chart

jsignell:
Haven't included it in any parts of the site yet.


jsignell:
Where should this be exposed on the site? In the intro or the talk maybe?

jbednar:
Given that there is a section called "Talks" (plural) that has only one talk, maybe add a new talk (i.e. a new notebook) about "Choosing viz tools", where we can put some intro text explaining what this is, and talking about how to use it?

PR #231, opened 2019-11-22
Updated mermaid diagram

jbednar:


jsignell:
I ended up just pushing to the branch since it was easier. Hope that is ok. I'll merge when green.

PR #234, opened 2019-12-05
Fix snappy support on travis

ceball:
Doesn't replace the value of deciding about #220, but fixes the failing travis builds. Would still require documentation like "pip users will need to install python-snappy's system dependencies using their system package manager". 

The successful pip/package test run is "test package on pip" at https://travis-ci.org/holoviz/holoviz/builds/621299758 (that usually only runs for cron/master - I temporarily enabled for this PR, then disabled again).

Note: python-snappy appears to have binary wheels for macos. Maybe we should ask if it would be possible to get wheels for all platforms? (Though I have no idea if the macos wheels contain all the necessary dependencies anyway...)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ env:
 
 stages:
   - name: test_user_install
+    if: type = cron  
   - test
   - name: package_dev
     if: tag =~ ^v(\d+|\.)+[a-z]\d+$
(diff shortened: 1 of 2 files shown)


ceball:
~~I should have marked this draft - don't merge yet even if the build passes.~~

jbednar (approved):

jbednar:
D'oh! Meant to squash merge, not rebase and merge.  Never mind...

PR #237, opened 2019-12-20
Remove testpath dependency

ceball:
testpath isn't in environment.yaml, so I removed it from setup.cfg too. Seems like whatever the original problem was might have gone away - you can see without testpath explicitly listed as a dependency, it's still installed (https://travis-ci.org/holoviz/holoviz/jobs/627770257#L974). The comment said it was an "unstated dependency of datashader".

Notes:
  * I also checked that generating the environment file using pyctdev results in the same environment file we currently have.


jbednar (approved):
Looks good to me!

PR #239, opened 2019-12-22
Replace channel 'https://conda.anaconda.org/pyviz' with 'pyviz'

ceball:
At some time, on some system (belonging to one of us - @jbednar or me maybe?), 'pyviz' alone did not work. I can't reproduce it on linux now, so maybe we could remove it and see what happens.

Also: Slightly refactor to be simpler to my mind (bit subjective); shouldn't have an impact on behavior.


jbednar (approved):
Looks fine to me; I can't think of any reason it should be fully specified like that anymore.

PR #242, opened 2020-02-17
Remove conda-forge from tests

jbednar:
(As it's not used in the installation instructions)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,6 +1,6 @@
 environment:
   global:
-    CHANS_DEV: "-c conda-forge -c defaults -c pyviz/label/dev"  # this order matters because we're using strict channel priority
+    CHANS_DEV: "-c defaults -c pyviz/label/dev"  # this order matters because we're using strict channel priority
     PIN: "--pin-deps"
   matrix:
     - PY: "3.6"
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,6 +1,6 @@
 environment:
   global:
-    CHANS_DEV: "-c defaults -c pyviz/label/dev"  # this order matters because we're using strict channel priority
+    CHANS_DEV: "-c defaults -c pyviz/label/dev -c conda-forge"  # this order matters because we're using strict channel priority
     PIN: "--pin-deps"
   matrix:
     - PY: "3.6"
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ os:
 
 env:
   global:
-    - CHANS_DEV="-c conda-forge -c defaults -c pyviz/label/dev"   # this order matters because we're using strict channel priority
-    - CHANS_REL="-c conda-forge -c defaults -c pyviz"  # this order matters because we're using strict channel priority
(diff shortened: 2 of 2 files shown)


ceball:
And this makes sense as a general move because we previously agreed we're generally testing "on top of defaults" on our CI, while testing "on top of conda-forge" would happen on conda-forge (although it doesn't necessarily happen yet).

Plus we'd hope there wouldn't be a big difference between the two anyway (though I'm not up to date on that at all - mainly I guess it's availability/version differences these days? i.e. not so many binary incompatibility or naming issues any more).

jbednar:
At the moment the issue was that the holoviz testing environment wasn't solving on conda-forge (https://travis-ci.org/holoviz/holoviz/jobs/650192940?utm_medium=notification&utm_source=github_status), and it seems fine on defaults now that I've switched. Michael Sarahan suggested that one problem is that our doit setup is first creating an empty environment and then populating it, and the solving process works better if it can do that all at once. Not sure how hard it would be to make that happen.  In any case, we don't really want _all_ at once; ideally we'd have the main dependencies set up in a single step, then separately install doc or testing dependencies so that those separate ones don't affect the initial solve except where absolutely necessary (otherwise we won't be testing a realistic approximation to user environments!).

ceball:
> Michael Sarahan suggested that one problem is that our doit setup is first creating an empty environment and then populating it, and the solving process works better if it can do that all at once.

Ok, as per step 6 and the tip below it at https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#creating-an-environment-with-commands ?

I guess it was set up the other way (create env, activate env, then install deps) because that's conceptually the same as for python itself. With python, you have to create and then activate an env before installing dependencies into it, as far as I know (e.g. https://docs.python.org/3.7/tutorial/venv.html).

> Not sure how hard it would be to make that happen.

It would be straightforward. But yes, it would require a change.

The next version of pyctdev (unreleased) already supports operating on named environments in general (with conda) rather than requiring the target environment to always be the active one.

> ideally we'd have the main dependencies set up in a single step, then separately install doc or testing dependencies so that those separate ones don't affect the initial solve except where absolutely necessary

That part is already supported. Instead of passing --extra x --extra y --extra z to one develop_install command (I forget the exact command and option names), you'd split into multiple such commands.

> (otherwise we won't be testing a realistic approximation to user environments!)

Anecdotal, but I have often seen people assume conda is broken and abandon it (for pip) in such scenarios (i.e. where they are trying to add something to an existing environment, but can't). Doing so is "wrong", but often works enough (at least "for now" - the problems are shifted to the future...).

jbednar:
Adding into an existing environment isn't wrong, but it does work differently from including that package at the start, because conda expects you to be happy with your installed packages, and only want to make minimal changes when adding this new one.  That's precisely the behavior we want for testing requirements -- don't disturb the existing environment when adding these new packages that I need just for testing. But it's not the behavior we want when initially setting up the environment; we want the requested packages and Python all at once, with no special preference for something like cffi that came along with Python. Sounds like that's possible ("straightforward") but someone would have to do it.

ceball:
I opened an issue on pyctdev. (I get the difference, and have no problem with it - was just explaining the reason for doing it the other way round.)

PR #266, opened 2020-06-02
CI updates

ceball:
Changes:
* Update windows CI to use python 3.7 (development install now gets further than it used to, but still not all the way through).
* Update mac CI to find python (after travis changes).
* Remove unused and out-of-date examples.yml (#267)
* Remove "unpinned"/"upcoming trouble" build - no longer possible as separation of "abstract" vs "concrete" dependencies has been removed from setup.cfg.
* The dependency `pyct[cmd]` for getting the examples is missing for pip. pyct the conda package is "fat" and includes pyct.build, pyct.cmd, etc, while pyct the pip package is minimal and includes just pyct.build. This change could...
  * be kept here (not sure how it will affect conda - at least it used to have no impact as the `[cmd]` part would be silently dropped by conda-build)
  * or be moved to e.g. panel
  * or be dropped/changed to documentation (not sure people will easily be able to do the tutorials via pip anyway...)

So it's not a great piece of work, but I think it doesn't make anything worse.

Things verified during intermediate commits in this PR:
  * released conda package works for running tutorial nbs on linux (via nbsmoke/not tested interactively)
  * released conda package works for running tutorial nbs on mac (via nbsmoke/not tested interactively)
  * released conda package works for running tutorial nbs on windows (via nbsmoke/not tested interactively)
  * the released pip package as is doesn't work for examples (at least also need: `pip install pyct[cmd]` to be able to _get_ examples; likely need more to be able to run examples - not yet tried)

The above should all become apparent when future cron jobs run.


jbednar (approved):
Thanks!

jbednar:
I'd merge, except for the failing build https://ci.appveyor.com/project/pyviz/holoviz/builds/33277503 (ruamel?).

ceball:
Hasn't every PR been merged recently with failing builds? That's what the start of my comment is referring to:

> Update windows CI [...] development install now gets further than it used to, but still not all the way through

Unless it's a different build you're talking about? A linux build failure would be a regression, yes...

ceball:
And I might investigate the windows build some more, but not tonight (and so probably best in a future PR) :)

jbednar:
@philippjfr indicated when I asked about that earlier today that the failing Appveyor builds were marked failing because they were cancelled, not because they actually failed. Sounds like that's not the case?

philippjfr:
Well that was definitely the case but it may also be the case that they were also actually failing. There was a huge backlog of builds so I didn't see it finish.

philippjfr:
Btw thanks Chris. Should have thought about the consequences of removing pyct like I did. Has anyone tested installing the 0.11.5 package on windows? Otherwise I can try in a bit.

ceball:
Ah, I just assumed from the red crosses that the appveyor "developer" builds were failing. If they weren't failing, then this PR would be a step back in its current state, so probably hold off merging (though the change is to move to python 3.7 as per the instructions).

> Has anyone tested installing the 0.11.5 package on windows? Otherwise I can try in a bit.

Yes, but only installation of conda package and then non-interactive nbsmoke running of tutorial notebooks (all passed, python 3.7). 

Such "user install testing" *should* be working on all platforms on ci via cron jobs (once this PR is merged to fix the mac and windows CI environments...unless the windows one was not actually broken with 3.6 and is with 3.7...).

ceball:
If I'd done it not on CI, I would have checked the lab extension instructions don't miss anything required i.e. that the notebooks actually work for me in lab :) But I didn't get that far yet.

philippjfr:
>Ah, I just assumed from the red crosses that the appveyor "developer" builds were failing.

They were failing so I don't think this is a step back. I was just in a rush and didn't wait for them to finish 😞  I'm going to push a change in channel priority to this PR and see if that fixes it.

philippjfr:
Changing the channel priority seems to have fixed it. I'll merge once both builds are passing.

PR #281, opened 2021-06-15
Converting tutorial to anaconda project and unpinning metapackage

jlstevens:
This should be a more robust approach moving forwards.


setup.cfg
```diff
@@ -31,22 +31,22 @@ include_package_data = True
...
 install_requires =
-    numpy ==1.18.1
-    pandas ==1.0.3
-    dask[complete] ==2.17.2
-    xarray ==0.15.1
-    matplotlib ==3.1.3
-    notebook >=5.5
+    numpy
+    pandas
+    dask[complete]
```
jbednar, line 36: Any idea whether we really need "complete"?

jbednar (approved):
Looks good! Will also need some instructions updated, of course.

setup.cfg
jlstevens: Honestly, I'm not sure...I simply removed the pins without thinking about anything else.

jlstevens:
@jbednar I've got a first cut of what should be working instructions for a project based approach (which you can view [here](https://github.com/holoviz/holoviz/blob/f256e42dddf05630e225994c710c20471428fce1/examples/tutorial/00_Setup.ipynb)).

Three sticking points right now:

1) Needing to pin fsspec to load the parquet file (now considered invalid, was working before). The best fix is to update the file but I don't yet understand what fsspec is complaining about (or how to fix it)
2) For images (mainly logos) to load in the notebooks, the server needs to be running one directory above the `tutorial` directory which means users need to navigate to the directory with all the notebooks. Assets could be moved or maybe we could risk a symbolic link?
3) I made the test for the 'min' workable bokeh, holoviews and datashader version simply be the currently locked versions. Is this ok? I'm sure some older versions could also run the tutorial...

jlstevens:
I'll merge this for now though we will still need to address the points I made above.

Issue #275, opened 2020-09-21
To-do list


jbednar:
Subsumes #26, #174, and #202, listing the remaining open to-do items about the tutorials.

- Review the existing tutorial content:
  * [ ] Are there any broken inter-notebook links? At some point there were some that went nowhere.
  * [x] Are there any warnings generated in the notebooks?
  * [ ] Review tutorial in the context of the website (as opposed to Jupyter)

- Improve the existing tutorial content:
  * [ ] Improve overall "finish" of the examples to make them look and feel awesome (see #260)
  * [x] Improve the earthquake plot from tutorial front page and 05_Composing_Plots (see #260)
  * [x] Fix logz option in earthquake plot in 05_Composing_Plots (see #260)
  * [x] Add a greyed out watermark or other unmissible indication for interactive content not interactive on the website (see #260)
  * [x] Simplify 01_Overview so that an inexperienced person can understand and appreciate it. Perhaps replace with 1-4min video showcase of solid, compelling, fast examples?
  * [ ] Bring "Workflow from data to decision" content content front and center (see https://github.com/holoviz/holoviz/issues/260#issuecomment-672706218)
  * [x] For simplicity, avoid using dask (i.e to avoid `.compute()` calls) unless it matters for performance
  * [x] Update examples to use rasterize() + dynspread instead of datashade() + dynspread
  * [x] Improve mpl examples, e.g. with `%matplotlib widget` and showing more advanced features
  * [x] Avoid the need for `python-snappy` which cannot be pip installed. See #220
  * [x] Make Exercise 1 not refer to material not yet introduced (explicitly creating widgets)
  * [x] small_df causes Bokeh hiccups in tutorial 4; make smaller?  Or limit hover?
  * [x] Fix for where the hover tool was being overridden in `.opts`. (where?)
  * [ ] If giving a 2 hour version like acon20.ipynb again, either should start with Overview.ipynb or Panel's tutorial and add exercises, or should add another notebook after basic plotting that brings Panel and hvPlot together in a simple way.
  
- Possible new tutorial content:
  * [ ] Boost coverage of libraries other than Bokeh, e.g. other .plot API examples from other libraries
  * [ ] Discuss how other tools fit in? Seaborn, ggpy, etc.
  * [ ] Explain what these packages will _not_ do
  * [ ] Add more links for further study -- having seen this one example, e.g. interact, where can we find all the details?
  * [ ] Add list of references and weblinks for further exploration
[...]

PR #283, opened 2021-06-21
Scipy 2021

jlstevens:
This branch should be where we update our tutorial for Scipy 2021.

Note that I am using the environment defined by the new `anaconda-project.yml` added in https://github.com/holoviz/holoviz/pull/281. In this first commit, I've simple reordered the material so that plotting is introduced before panel. My next step is to integrate these sections in a way that makes more sense i.e to use the composite plots built with `hvplot` in the panel section instead of only showing simple dataframe tables and logos.


jbednar:
To-do items remaining after my edits:

## Immediate
- [x] Fix file paths to make it servable (JL)
- [x] Remove snappy requirement (https://github.com/holoviz/holoviz/pull/220)? (JL)
- [x] Project data ahead of time so that we can talk about projections only in an aside (explaining for those interested), and not have to do it in real time. Most users probably don't need to spend brainpower thinking about projection. (JL)
- [x] Remove dask, for simplicity (JL)
- [x] Pin bokeh 2.3.3 to avoid cnorm='eq_hist' colorbar bug (JL)
- [x] Make hv dev release

## General
- [x] Set up binder (https://github.com/holoviz/holoviz/pull/206) or provide AE sessions?
- [x] Make sure the overview of Panel emphasizes that it is general (not tied to a plotting library), plus components, pn.bind, layouts, .show(), serve, etc., motivated each time by what they let the user do (Jim)
- [x] Review https://github.com/holoviz/holoviz/issues/275 looking for items to do or that are no longer relevant (Blythe)
- [x] Put content cut from https://github.com/holoviz/holoviz/pull/284 somewhere, e.g. in a wrapup notebook at the end (Jim)
- [x] all: Update inter-notebook links in the last section of each page
- [ ] Update website (preferably with the new sphinx theme)
- [ ] Run a link checker on the final version

# Exercises (should all be .ipynb, not .py)
- [x] exercise 1
- [x] exercise 2
- [x] exercise 3
- [x] exercise 4
- [x] exercise 5

## 00
- [x] 00_Setup.ipynb and 01_Overview: Update screenshot 
- [x] 00_Setup.ipynb: Update time estimates
- [x] 00_Setup.ipynb: Add links to updated exercises

## 02
- [x] Replace `datashade=True` with `rasterize=True, cnorm='log'` (Jim)

## 03
- [x] 03_Composing_Plots: What should we see with "Use tab completion to explore weekly_count_plot."? The hint says to look at .data, but I don't see anything informative there. (Jim)

## 04
- [x] 04_Interlinked_Plots: Why do you do `ls(depth_hist)+ ls(mag_hist)` instead of `ls(depth_hist+ mag_hist)`? (Jim)
- [ ] 04_Interlinked_Plots.ipynb: For Linked-brushing-across-element-types, the map plot aspect ratio is broken when I do a box selection, and lasso selection seems to have no effect.

## 05
- [x] 05_Interactive_Pipelines.ipynb: Fix "TODO: hangs without print"
- [ ] 05_Interactive_Pipelines.ipynb: Filtering-the-earthquakes-on-a-map flickers a lot!
- [x] Overall, this one seems confusing. I think we need a few self-contained .interactive examples before having anything spread over multiple cells.

## 06 Large Data
- [x] Remove and renumber what comes after it.

## 06

- [x] Finish notebook
- [x] Talk about the limitations of .interactive, .hvplot, etc. -- how do users know when they hit the limits of the magic? What do they do then?
- [ ] The final app is not very responsive. :-(

## 08

- [x] The last example does not seem to update at all when widgets are used

## Later cleanup
- [ ] Make final metapackage release (and remove package-building machinery from this repo?)
- [ ] Add repr for bound objects (as a separate pn.show?) amd add param.bind
- [x] Fix hvPlot for dynspread=True, rasterize=True (see https://github.com/holoviz/hvplot/pull/637)
- [x] Release dev package for hvplot with https://github.com/holoviz/hvplot/pull/637

ppwadhwa:
I quickly went through the tutorial, mostly looking that it made sense. I really like it!  It flows well and the story you are telling is clear to me.

PR #284, opened 2021-06-22
Revise intro

ppwadhwa:
The introduction needs to be reduced to 10 minutes.  I cut out some chunks to shorten it.  

A few comments:
I left in the visuals that show what Holoviz covers, with 1 coming right after the other (in the case of what holoviz covers and then what is most supported).  It feels a bit weird to have them both right next to each other in cells without text in between, but I also don't know that there is much to say beyond what is there.  If they were slides, I would put them 1 after the other.  

I also left a bunch of stuff at the bottom, because I think its nice information to have, but its not necessarily needed in the intro.  So, my suggestion is either some supplementary material, or have it at the end as a footnote.  I'm not sure.  I just wanted to show which parts I wanted to keep somewhere.

### DIFF
--- a/examples/tutorial/01_Overview.ipynb
+++ b/examples/tutorial/01_Overview.ipynb
@@ -80,22 +80,24 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Shortcuts, not dead ends\n",
+    "## HoloViz libraries\n",
     "\n",
-    "<img src=\"../assets/shortcuts.png\" width=70% align=\"left\">"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "## HoloViz principles\n",
+    "<img width=\"800\" src=\"../assets/pn_hp_hv_gv_ds_pa_cs.png\"/>\n",
     "\n",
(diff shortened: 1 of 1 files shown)


jlstevens:
Having talked to Pam about this today, my comments are:

* I agree that cutting out the 'shortcuts, not dead ends' section will shorten things nicely but maybe we still want a bullet point to mention it and maybe a link to a page explaining that principle.

* The 'built on scientific python ecosystem' section also makes sense to move out of the main introduction. Again, maybe one bullet point and a link to somewhere else instead?

* Getting rid of the summary also makes sense: the intro/overview should be short enough not to need a summary!

* The 'Getting started' links should go to the end of the tutorial to show people all the other things they can do. Linking to all those possibilities at the start of the tutorial is distracting.

ppwadhwa:
@jbednar Jean-Luc and I went through the tutorial introduction and talked a bit about the "shortcuts not dead ends" part.  We agree that its a bit of a distraction for a short introduction.  We also think its important to mention somewhere.  Maybe at the end to wrap up the tutorial?  Thoughts?

jlstevens:
Thanks @ppwadhwa for tightening this up. I'm going to merge to master then rebase into the main PR.

PR #292, opened 2021-07-19
Switch from Travis CI to GitHub Actions

philippjfr:


jlstevens:
Other than updating the tutorial TOC, the dev website already looks good! 

The main thing that could have been outdated was the installation instructions but those were already updated for this tutorial.

A few notes:

* I've added entries for the three notebooks at the end of the tutorial into the TOC but commented them out as this material is explicitly marked optional (the links from the index page do work though). 
* We could review the roadmap to see if there is anything we want to update for 2021.
* I was tempted to remove the talk but have opted not to because 1) it looks like it all still works 2) it mostly covers `.hvplot` which is still a suitable entry point. @jbednar if you want to change anything about the talk (perhaps just the very end that still uses `interact`?) then we can easily rebuild the site.

jlstevens:
Seems to be working well! Merging.

PR #295, opened 2021-08-03
Replace earthquake data with downsampled version on binder

philippjfr:
Previously it wouldn't overwrite the larger file so you had to manually rename to get that to work.


jbednar:
Thanks! I guess it wasn't feasible to simply download the smaller one instead, which would presumably save time when setting up the binder instance?  E.g. by setting some environment variable that anaconda-project would respect?

philippjfr:
Honestly doesn't seem worth it since the binder image is cached.

PR #297, opened 2021-08-19
Remove .compute()

maximlt:
Because it raises an error


jbednar:
Good point! I think that's left over from when the example used Dask, but it no longer appears to and thus should not call .compute().

PR #299, opened 2021-09-06
Make dashboard notebooks into python files

ppwadhwa:
I have made python files from some of the dashboard notebooks from the holoviz tutorial (from notebooks 06, 07, and 09.  08 didn't seem like it warranted a separate python file).  I didn't include comments, because I wasn't sure if that was needed.  I can add them if you all would prefer that.

These work for me as is (i.e. `panel serve 06_Dashboards.py`).  Let me know what you think!

### DIFF
--- a/examples/tutorial/06_Dashboards.py
+++ b/examples/tutorial/06_Dashboards.py
@@ -1,0 +1,67 @@
+import pandas as pd
+import panel as pn
+import holoviews as hv
+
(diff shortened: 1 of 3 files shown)


jbednar:
Nice! Those look great. I'll go ahead and merge them, then we can separately add links to them from the notebooks. Thanks so much!

PR #301, opened 2022-04-04
Update the tutorial's dependencies

maximlt:
In preparation of the tutorial for PyConDE/PyData.


examples/talks/Overview.ipynb
```diff
@@ -2,11 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {
-    "slideshow": {
-     "slide_type": "skip"
-    }
-   },
```
jbednar: This metadata is important; can you please restore all of it for this notebook? This notebook works as a slideshow (or at least it used to), which it can't do without that metadata. We should update the jq stripping command to ignore slideshow metadata.

examples/tutorial/02_Plotting.ipynb
```diff
@@ -60,7 +60,6 @@
    "source": [
     "%%time\n",
     "df = pd.read_parquet('../data/earthquakes-projected.parq')\n",
-    "df.time = df.time.astype('datetime64[ns]')\n",
```
jbednar, line 63: Has the data file been updated to no longer require this? Or does HoloViews now handle it better?

examples/talks/Overview.ipynb
maximlt: Oh oh I had missed that, thanks! I reverted it in https://github.com/holoviz/holoviz/pull/301/commits/fdec9e1a0440985432b598fde9e711cbe97e0183

examples/tutorial/02_Plotting.ipynb
maximlt: Difficult for me to tell what was the problem before, but with new versions of pandas this line was raising a warning. Removing it caused no problem as far as I could tell.

jbednar: Hmm. Very suspicious! @jlstevens and @philippjfr , any idea why that was previously required?

jlstevens: Can't remember, sorry! But if that line was added it was probably as a workaround that may not longer be needed (if indeed there are no problems afterwards)

maximlt: I've ran the notebooks locally without any problem after this change. The test suite also passes (except on Windows, but that's another error). There's always a small chance that I missed something...

PR #306, opened 2022-06-23
Updates with matplotib/plotly and function as input

maximlt:
cc @jbednar 

This includes additional sections on:
- the new matplotlib/plotly support
- using a function as input of an interactive pipeline


examples/tutorial/02_Plotting.ipynb
```diff
@@ -563,7 +563,105 @@
    "source": [
     "Here it is clear that the most commonly detected events are light, and typically shallow.\n",
     "\n",
-    "# Exploring further\n",
+    "## Output Matplotlib or Plotly plots\n",
+    "\n",
+    "While the default plotting backend/extension of hvPlot is Bokeh, it also supports rendering plots with either [Plotly](https://plotly.com/python/) or [Matplotlib](https://matplotlib.org/). As Bokeh, Plotly is an interactive library that provides similar data exploring actions (pan, hover, zoom, etc.). You can decide to use Plotly instead of Bokeh if you prefer its look and feel, or if your organization's policy is to use that plotting library. Matplotlib is mostly a static plotting library, it is great at creating static images (e.g. PNG) that you can then easily save and share with your friends or coworkers, and embed in documents for publication.\n",
```
jbednar:
```suggestion
    "While the default plotting backend/extension of hvPlot is Bokeh, it also supports rendering plots with either [Plotly](https://plotly.com/python/) or [Matplotlib](https://matplotlib.org/). Like Bokeh, Plotly is an interactive library that provides similar data-exploring actions (pan, hover, zoom, etc.). You can decide to use Plotly instead of Bokeh if you prefer its look and feel, or if your organization's policy is to use that plotting library. Matplotlib is supported for static plots only (e.g. PNG or SVG), which is useful for saving and sharing images and embedding them in documents for publication.\n",
```

examples/tutorial/02_Plotting.ipynb
```diff
@@ -563,7 +563,105 @@
...
+    "hvplot.extension('plotly', 'matplotlib')\n",
+    "\n",
+    "small_df.hvplot.scatter(x='longitude', y='latitude')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Once a backend is loaded with `hvplot.extension` you should use the `hvplot.output` function to switch from a backend to another."
```
jbednar:
```suggestion
    "Once a backend is loaded with `hvplot.extension` you can use the `hvplot.output` function to switch from one backend to another."
```

examples/tutorial/02_Plotting.ipynb
```diff
@@ -563,7 +563,105 @@
...
+    "You can easily save a plot with the `hvplot.save` function, which is particularly useful when you are using Matplotlib as your plotting backend."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "hvplot.save(plot, 'plot.png')"
```
jbednar: Can we also show SVG? I think that will address people who think these tools aren't for publication.

examples/tutorial/02_Plotting.ipynb
```diff
@@ -563,7 +563,105 @@
...
+   "source": [
+    "from IPython.display import display, Image\n",
+    "display(Image('plot.png'))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "While hvPlot allows you to customize your plots quite extensively there are always situations where you want to customize them a little more than what can be done directly with hvPlot. In those cases you can use the `hvplot.render` function to get a handle on the underlying figure object. Below we're using Matplotlib's `xkcd` context manager to turn on *xkcd* sketch-style drawing mode during the render of the figure."
```
jbednar:
```suggestion
    "While hvPlot allows you to customize your plots quite extensively there are always situations where you want to customize them a little more than what can be done directly with hvPlot. In those cases you can use the `hvplot.render` function to get a handle on the underlying figure object. Below we're using Matplotlib's `xkcd` context manager to turn on *xkcd* sketch-style drawing mode while rendering the figure."
```

examples/tutorial/02_Plotting.ipynb
```diff
@@ -563,7 +563,105 @@
...
+    "\n",
+    "print(mpl_fig)\n",
+    "mpl_fig"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "`mpl_fig` is a Matplotlib `Figure` object that we could further customize using Matplotlib's API."
```
jbednar:
```suggestion
    "`mpl_fig` is a Matplotlib `Figure` object that we could further customize using [Matplotlib's API](https://matplotlib.org/stable/api)."
```

examples/tutorial/05_Interactive_Pipelines.ipynb
```diff
@@ -311,6 +311,54 @@
     "For readability, seven columns were chosen before displaying the `DataFrame`. Have a look at `df.columns` and pick a different set of columns for display."
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Functions as inputs\n",
+    "\n",
+    "Quite often the data structure you want to explore in a pipeline, e.g. a Pandas DataFrame, may itself be the outcome of another pipeline. It may for instance be extracted and transformed from a database or an API call, or the output of some simulation or pre-processing. With `hvplot.bind` you can bind an input function that returns the data structure you want to explore, to widgets that when updated will re-execute that function. \n",
```
jbednar:
```suggestion
    "Quite often the data structure you want to explore in a pipeline, may itself be the outcome of another pipeline. It may for instance be a Pandas Dataframe created by extracting and transforming the output of a database or an API call, or it could be the dynamic output of some simulation or pre-processing. With `hvplot.bind` you can start with an arbitrary custom function that returns the data structure you want to explore and then bind that function's argument to widgets. Then when those widgets change, the function will get called to get the updated output.\n",
```

examples/tutorial/05_Interactive_Pipelines.ipynb
```diff
@@ -311,6 +311,54 @@
...
+    "\n",
+    "For the sake of simplicity we will not fetch data from an external location, instead we create a simple function that filters the earthquakes dataset by event type and returns a DataFrame."
```
jbednar:
```suggestion
    "To keep this example self-contained we'll illustrate this process using a simple function that filters the earthquakes dataset by event type and returns a DataFrame, but this function could include _any_ computation that returns a DataFrame."
```

(but it seems like a more canonical example would be for a user to select one of several CSV files, which _requires_ an approach like this and is a very common need.)

examples/tutorial/05_Interactive_Pipelines.ipynb
```diff
@@ -311,6 +311,54 @@
...
+   "source": [
+    "def input_function(event_type):\n",
+    "    return df[df['type'] == event_type]"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We can then create a Panel `Select` widget with a few options, widget that we bind to `input_function`. Calling `.interactive()` on the bound object is what allows it to be used in an interactive pipeline, as we previously did with `dfi`."
```
jbednar:
```suggestion
    "We can then create a Panel `Select` widget with a few options and bind it to the `input_function`. Calling `.interactive()` on the bound object is what allows it to be used in an interactive pipeline, as we previously did with `dfi`."
```

jbednar (approved):
Looks great, thanks! Minor tweaks suggested.

examples/tutorial/02_Plotting.ipynb
maximlt: I've added an example with a SVG output.

PR #313, opened 2022-07-27
Anaconda project 0.11.0 compatibility

### LINKED ISSUE #312: Update anaconda-project.yml and anaconda-project-lock.yml with release of new anaconda-project
With the release of [`anaconda-project` v0.11.0](https://github.com/Anaconda-Platform/anaconda-project/releases/tag/v0.11.0) the current lock file fails to solve because it now ignores channels defined in the .condarc file.

 </img>

Furthermore, support for `macos-arm64` has been added, which should be added to the project and the comments about `export CONDA_SUBDIR=osx-64` should be removed. 

 <s>When ran `anaconda-project update` it failed on docs environment on `linux-64` with the following message:

 </img>

Looking at the error message and the anaconda_project file it is because the packages are not wrapped in quotes. If I add a space between the package and version, `anaconda-project update` is successful ran. This seems to be a regression with the new release of `anaconda-project`. I will file an MRE bug for anaconda-project.</s> Can't seem to recreate the bug anymore so will not do anything.

https://github.com/holoviz/holoviz/blob/70e0ea4a5d24cb4b399a224e5384cf0f93feb54d/examples/anaconda-project.yml#L61-L65

hoxbro:
Fixes #312

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -45,7 +45,7 @@ twice to get the very latest) to ensure you have the latest version::
 
 You should then make sure you have `anaconda-project` installed::
 
-   > conda install anaconda-project
+   > conda install "anaconda-project>=0.11"
 
 
 
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -43,10 +43,15 @@ twice to get the very latest) to ensure you have the latest version::
    > conda update -n base conda
    > conda update -n base conda
 
-You should then make sure you have `anaconda-project` installed::
+We recommend installing `anaconda-project` is a new environment, to keep
+things separate and clean. We'll create an environment called `project`, 
+please select another name if you already have an environment named this way::
 
-   > conda install "anaconda-project>=0.11"
+   > conda create -n project "anaconda-project>=0.11"
 
+Now activate this environment::
+
+   > conda activate project
 
 
 Step 2: Download and extract the tutorial project
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -47,10 +47,6 @@ You should then make sure you have `anaconda-project` installed::
 
    > conda install anaconda-project
 
-If you're on a M1 Mac you'll currently need to direct it to use
-emulated binaries (but skip this step if not on an M1 Mac)::
-
-   > export CONDA_SUBDIR=osx-64
 
 
 Step 2: Download and extract the tutorial project
--- a/examples/anaconda-project-lock.yml
+++ b/examples/anaconda-project-lock.yml
@@ -17,10 +17,11 @@ locking_enabled: true
 env_specs:
   test:
     locked: true
-    env_spec_hash: 8e77a647eaedf8158c733d7546f21e73a23cafb0
+    env_spec_hash: 61ef260560c348328aca81ca22579a3fc3d252ca
     platforms:
     - linux-64
     - osx-64
+    - osx-arm64
     - win-64
     packages:
       all:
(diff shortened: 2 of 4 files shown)


jbednar:
What will happen when someone uses an older anaconda-project version now?  Unless we demand that they install a particular version, I don't think we can assume they have 0.11.

hoxbro:
This works on the previous versions too. I have tested it. The newer version is more strict in what it accepts, so if version 0.11.0 is installed the current lock file will fail. 

With version 0.11.0 now on the [defaults channel](https://anaconda.org/main/anaconda-project), a new user following the [installation guide](https://holoviz.org/installation.html) will get version 0.11.

jbednar:
Great! However, what about under M1 Macs? Without those instructions, seems like it will still fail for older anaconda-project versions. Shouldn't the instructions still include that bit but say `(not needed if you have anaconda-project 0.11 or later)`?

maximlt:
Why not just suggesting to install `anaconda-project >= 0.11`? In which case `anaconda-project` will be updated if the user already has it installed.

An unrelated comment, at the tutorial there was someone with a broken `base` environment. I usually try to install as few things as possible into `base`, I think it's a recommended practice and I'm pretty sure having read somewhere that `base` might one day become a protected environment, to avoid this kind of issue. In this case the workaround was easy, just install `anaconda-project` in its own environment and activate it. I'd vote for updating the docs with this additional step.

doc/installation.rst
```diff
@@ -43,14 +43,15 @@ twice to get the very latest) to ensure you have the latest version::
    > conda update -n base conda
    > conda update -n base conda
 
-You should then make sure you have `anaconda-project` installed::
+We recommend installing `anaconda-project` is a new environment, to keep
+things separate and clean. We'll create an environment called `project`, 
+please select another name if you already have an environment named this way::
 
-   > conda install anaconda-project
+   > conda create -n project "anaconda-project>=0.11"
```
jbednar, line 50: I'm not sure I agree with this step. I agree that people shouldn't normally install into their base environment, but we've never told anyone to do that. If they are defaulting to installing into base, I don't think it's up to us to worry about that; they'll have their own problems with that, and it's unlikely that anaconda-project would be the culprit. In fact, if there is anything that _should_ be in a base environment, it's meta-utilities like anaconda-project, which enable working with other environments. Rather than get into any of that, I'd rather keep these instructions simple and have people just install anaconda-project wherever it's convenient for them.

maximlt: I remember suggesting that change in a discussion and having no approval/no objection, and I'm not the one running the tutorial, so I'm fine reverting!

For reference, this change was motivated by someone I helped during the tutorial. That person had a broken *base* environment and basically the only solution was to create a new environment to install anaconda-project. It happens that it's also my current setup.

> it's meta-utilities like anaconda-project

As far as I know there's no concept of meta utilities in conda. I'd love to have something like `pipx` - that is now pretty popular in the pip world - to easily install utilities in their own environment and have them added to PATH.

> it's unlikely that anaconda-project would be the culprit

True. Yet it does come with a number of dependencies.

> I agree that people shouldn't normally install into their base environment, but we've never told anyone to do that

Well we don't tell them to activate any other environment, so they're very likely to install `anaconda-project` in their base environment. After all the tutorial is given by Anaconda employees, they have no reason not to trust us on our install procedures and decide to add an extra step. Probably some people in the room have heard that it's not good practice to install packages in their base environment, yet I'd bet they will still do it. Not providing more details on where and how to install anaconda-project assumes that people know what they're doing when they run `conda install`, that's a big assumption :)

I think an interesting effect of this change is that it leaves the system of those who follow the tutorial in a state that easily allows them to get back to their initial state. They can just decide to remove the environments that have been created; how to clean up what has been installed/downloaded  should actually be documented somewhere in the tutorial 🙃 Currently they'd have to remove anaconda-project from *base*, in all likelihood. Attempting to remove anything from *base* gives me goosebumps!

PR #325, opened 2022-09-11
simplify text on tutorial index page

droumis:
- [x] Make the intro text start by describing what the user will accomplish
- [x] Adjust the language around the list of Holoviz packages making it seem less like the tutorial will touch all of them
- [x] Remove the icons of the HoloViz packages, they are shiny distractions that take attention away from the dashboard
- [x] Remove discussion about HoloViz vs PyViz
- [x] Remove underscores in 'interlinked_plots' and 'interactive_pipelines'
- [x] Remove mention of timed 'breaks'... they make it seem like this self-paced online tutorial is not the primary audience for the material
- [x] Remove the Related Links section for now. Without guidance as to why the users should go to one or the other, they are a distraction that makes the page look busy

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -40,6 +40,8 @@ jobs:
       - uses: conda-incubator/setup-miniconda@v2
         with:
           miniconda-version: "latest"
+          mamba-version: "*"
+          channels: pyviz/label/dev,conda-forge,nodefaults
       - name: Fetch unshallow
         run: git fetch --prune --tags --unshallow -f
       - name: Set output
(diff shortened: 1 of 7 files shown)


droumis:
Before
holoviz org_tutorial_index html (2)

After
pyviz-dev github io_holoviz_tutorial_index html

jlstevens:
At a glance, this looks great!

I'll do a full pass on this later but here one thing that jumped out at me:

*In this tutorial, you will explore a dataset of earthquake events and create a dashboard like in the image below.*

I'm not sure how to phrase things yet, but what we want users to accomplish is a better idea of how to achieve visualization tasks, and as much as I like this dataset, the earthquakes aren't the important point here :-)

droumis:
> At a glance, this looks great!
> 
> I'll do a full pass on this later but here one thing that jumped out at me:
> 
> _In this tutorial, you will explore a dataset of earthquake events and create a dashboard like in the image below._
> 
> I'm not sure how to phrase things yet, but what we want users to accomplish is a better idea of how to achieve visualization tasks, and as much as I like this dataset, the earthquakes aren't the important point here :-)

Ok I'll take another crack at it soon

PR #333, opened 2022-10-24
CI: fix the docs build

maximlt:


droumis:
is it weird that there are no packages pulled from the conda-forge/label/broken channel in the (now working) CI build, even though that was the only change here?

maximlt:
Oh no in this PR I've not added the conda-forge/label/broken channel, I've just re-formatted the list of channels from a single-line to a multiple-line format. The big change is the new lock file, that seems not to break the test suite and to unblock the docs build, the dev website built from this PR being live now at https://pyviz-dev.github.io/holoviz/contributing.html.

The contributing guide has some issues with the hints (they all appear nested), I must have done something wrong. I also think it should actually be broken down into a multiple pages:
- The first two sections (why? how? and developer instructions) should go into one page, that we can link to from other sites
- The operating guide should have its own page
- There should be a page about how to contribute to holoviz.org itself

On that last item I'll just write down here some quick notes for now:
- create an environment with `anaconda-project` (I don't like installing stuff in the base environment, that's a recipe for a disaster)
- `cd` in to the `/examples` folder
- run `anaconda-project prepare --all` to prepare the project (this will download the datasets, and create all the environments declared in the project)
- run `anaconda-project lint` to run the linters
- run `anaconda-project test` to run the tests
- run `anaconda-project lock` to lock the project again (not sure if the lock file has to be manually deleted before, or if there's a flag to override it. Also it may actually be wiser to lock first and then run `prepare` as doing it the other way around may end up with different environments? Maybe @jlstevens you could help us with some anaconda-project best practices, I always get confused with its workflow)

Notes on how to build the sites should also be added to the contributing page of holoviz.org.

maximlt:
I think I'll also add the triaging guide lines to the operating guide, having everything at one place is better than scattered across sites and platforms. The triaging meeting and its hackmd link are already mentioned in the guide.

cc @jbednar

droumis:
> Oh no in this PR I've not added the conda-forge/label/broken channel, I've just re-formatted the list of channels from a single-line to a multiple-line format.

ahh yea sorry, it looks like I was looking at a [previous commit](https://github.com/holoviz/holoviz/pull/333/commits/133bc180036036793b39eff70cd0b2585a46c57d) to this branch

jlstevens:
> Maybe @jlstevens you could help us with some anaconda-project best practices, I always get confused with its workflow

Sure, happy to discuss this. That said, it might be time to start using (or at least try out) `conda project` or at least `conda lock` when it comes to locking...

PR #344, opened 2022-12-22
adds calendar of meetings to community page

droumis:
image

### DIFF
--- a/doc/about.rst
+++ b/doc/about.rst
@@ -93,6 +93,6 @@ is the Maintainers, one of which is the Project Director, and another may be the
 Lead Maintainer. The Project Director is responsible for the overall direction and 
 scope of the Project while a Lead Maintainer is responsible for releases and 
 day-to-day technical management of the Project. Read more about the 
-`Project Governance <https://github.com/holoviz/holoviz/blob/master/doc/governance/project-docs/GOVERNANCE.md>`_, 
+`Project Governance <https://github.com/holoviz/holoviz/blob/main/doc/governance/project-docs/GOVERNANCE.md>`_, 
 or find the governance for other HoloViz Projects in the appropriate 
 repository within the `HoloViz GitHub Organization <https://github.com/holoviz>`_.
--- a/doc/about.rst
+++ b/doc/about.rst
@@ -92,7 +92,7 @@ let's consider this HoloViz.org website Project. The governing body of this Proj
 is the Maintainers, one of which is the Project Director, and another may be the 
 Lead Maintainer. The Project Director is responsible for the overall direction and 
 scope of the Project while a Lead Maintainer is responsible for releases and 
-day-to-day technical management of the Project. Read more about this website's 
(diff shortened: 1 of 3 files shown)


doc/community.rst
```diff
@@ -50,15 +50,32 @@ details. Usually on gitter our response will just be "please open a
 Discourse issue". Gitter is well suited for chatting and live
 discussion, but because it is not archived and not easily searchable,
 we discourage using it for anything else.
+
+**Want to become a contributor?** We hold recurring developer meetings 
+and welcome anyone interested in becoming a contributor.
+
+  - General HoloViz meeting: This occurs biweekly on Fridays and is the best option for new members to join.
```
maximlt: Could you link to the meeting notes? I think they're all under the same HackMd now, if so that would be just one link above or below this list. I think that's useful for people to see what is usually being discussed.

maximlt: Somewhat unrelated to this PR but actually list items in Markdown should not be indented, when they are they're parsed as a block or quote (I'm not sure). See how it looks now https://holoviz.org/community.html

image

It'd be nice if you could fix those too :)

maximlt (changes requested):
Oh that's going to be nice!

Made some minor comments. If you feel like I think there are a few changes to be made to this page that could improve it:
- It feels like it has quite a lot of text. For instance I prefer [Dask's page](https://docs.dask.org/en/stable/support.html). So I'd say feel free to remove anything that you find superfluous :)
- The page recommend first to have a look a StackOverflow. I think our updated recommendation would strongly steer users towards Discourse.
- It's missing Panel's twitter handle and it'd be nice to turn these handles into proper links.

One last thing, the *Operating Guide* also mentions these meetings (not the docs one actually). I guess that should be mentioned in one place only and the *Community* page feels like the right one, could you replace mentions of these meetings in the *Operating Guide* by a link to the *Community Page*?

droumis:
@maximlt, I reworded and simplified most of the community page. If you like it we can merge now

doc/about.rst
```diff
@@ -92,7 +92,7 @@ let's consider this HoloViz.org website Project. The governing body of this Proj
 is the Maintainers, one of which is the Project Director, and another may be the 
 Lead Maintainer. The Project Director is responsible for the overall direction and 
 scope of the Project while a Lead Maintainer is responsible for releases and 
-day-to-day technical management of the Project. Read more about this website's 
-`Project's Governance <https://github.com/holoviz/holoviz/blob/main/doc/governance/project-docs/GOVERNANCE.md>`_, 
+day-to-day technical management of the Project. Read more about the 
+`Project Governance <https://github.com/holoviz/holoviz/blob/master/doc/governance/project-docs/GOVERNANCE.md>`_,
```
maximlt:
```suggestion
`Project Governance <https://github.com/holoviz/holoviz/blob/main/doc/governance/project-docs/GOVERNANCE.md>`_, 
```

maximlt (approved):
I really like the intro of the Community page 👍

PR #362, opened 2023-04-19
HEP 1: Python version support

ianthomas23:
This PR adds the first HoloViz Enhancement Proposal, on python version support, to the holoviz repo. It is placed in a new `hep` directory, and the idea is that future HEPs will be placed in the same directory and there will be one or more links from holoviz.org to them.

Rendered markdown is at https://github.com/holoviz/holoviz/blob/hep1/hep/hep1.md.

Everything here is open to discussion: the HEP itself, the format of the HEP, the directory structure, etc.

### DIFF
--- a/hep/hep1.md
+++ b/hep/hep1.md
@@ -1,0 +1,105 @@
+# HEP 1: Python version support
+
+This enhancement proposal describes the possible policy that HoloViz projects
+might adopt for support of Python versions. It is currently a discussion
+document and has not yet been accepted.
+
+## Background
+
+Historically a new version of Python has been released every one to two years,
+and in future annual releases are planned every October. The lifespan of each
+version is typically five years, so on any particular date there are expected
+to be five active Python versions. The current status of Python versions is
+illustrated in the
+[Python Developer's Guide](https://devguide.python.org/versions).
+
+NumPy have a more restrictive policy of supporting Python version released in
+the last 42 months (three and a half years), with a few extra caveats, in
+[NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html). They
+therefore support the latest three to four Python releases.
+
+A number of projects in the scientific Python ecosystem follow the NumPy policy
+including
+
+- [Matplotlib](https://matplotlib.org/stable/devel/min_dep_policy.html)
+- [numba](https://numba.readthedocs.io/en/stable/release-notes.html#version-0-47-0-jan-2-2020)
+- [pandas](https://pandas.pydata.org/pandas-docs/stable/development/policies.html)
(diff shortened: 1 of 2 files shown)


droumis:
looks good. Might be good to mention this in the SC meeting next Friday April 28

jbednar (approved):
LGTM

ianthomas23:
This has 2 approvals. I'll give people until Monday to express any contrary opinions, otherwise I will merge it myself.

hep/hep1.md
```diff
@@ -0,0 +1,105 @@
...
+Colorcet and Param currently support Python 3.7 upwards. When this reaches end
+of life at the end of June, the minimum support Python version will increase to
+3.8 for their next releases.
+
+The current Bokeh release (3.1.0) supports minimum Python of 3.8 and the next
+release (3.2.0) will support minimum Python 3.9. All HoloViz projects should
+drop support for Python 3.7 as soon as possible to support a minimum Python of
+3.8. Following release of Bokeh 3.2.0 there should be one further release per
+HoloViz project that supports Python 3.8 and then the minimum Python version
+for each should increase to 3.9.
```
maximlt, line 98: > Following release of Bokeh 3.2.0 there should be one further release per
HoloViz project that supports Python 3.8 and then the minimum Python version
for each should increase to 3.9.

That does not apply to Param and Colorcet, given that they're in the group 1 you listed above?

maximlt (commented):
The document defines the policy for when a Python version should be dropped. All the HoloViz projects (attempt to) follow Semantic Versioning. I *think* that our convention for dropping support for a Python version was that it wouldn't happen in a *patch* version but at least in a *minor* version. Now, I'm not entirely sure that was the convention, and if it was, I'm not entirely sure we should keep following it. And really, I have no opinion on that, except that I think we should agree on what's acceptable and probably add it to that document?

hep/hep1.md
ianthomas23: Correct. I'll add extra clarification, something like `All HoloViz projects (except Colorcet and Param which are covered above) ...`

ianthomas23:
@maximlt I'm not familiar with the (possible) convention on relating project and python versions, either within HoloViz or outside. If there is a convention going forward then yes we should add a comment or link here about it. But probably the best approach would be for the HoloViz project versioning to be documented first, e.g. in HEP2, even if it is just a single paragraph linking to standard SemVer practice.

hep/hep1.md
ianthomas23: I have added extra clarification here.

maximlt: Thanks!

PR #367, opened 2023-06-25
Update project

hoxbro:
Update the versions in anaconda-project, and then I ran `anaconda-project update`.

### DIFF
--- a/examples/anaconda-project-lock.yml
+++ b/examples/anaconda-project-lock.yml
@@ -17,7 +17,7 @@ locking_enabled: true
 env_specs:
   test:
     locked: true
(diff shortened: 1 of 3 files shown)


maximlt:
In case there was a doubt, feel free to self-merge this when you're happy with it as it might be that some other tutorial updates depend on it.

hoxbro:
Thank you. I just wanted to run through the notebooks and see if there are any glaring problems. I have already found one in pyviz_comms that is using an old version, `pyviz_comms=2.0.0a2=py_0`. 

Also, I want to note that notebook 8 probably should be updated with the new panel reactive features.

PR #374, opened 2023-07-10
Website updates

maximlt:
- Removed the link to the `Introduction` (home) page in the navigation bar, instead you can just click on the logo to get back to this page
- Removed some old CSS and templates, they should all be inherited from `nbsite` now, and update it to its latest version
- Turned off some matplotlib warnings due to missing xkcd fonts (https://github.com/matplotlib/matplotlib/issues/26282) => I tried getting rid of those in a second round but failed partially, so I re-filtered the warnings
- Simplified the tests and docs workflows
- Remove old packaging cruft
- Added link to the `Blog` in the navigation bar


doc/conf.py
```diff
@@ -35,11 +34,14 @@
             "url": "https://discourse.holoviz.org/",
             "icon": "fab fa-discourse",
         },
-    ]
+    ],
+    "header_links_before_dropdown": 15,  # no dropdown menu basically
```
droumis: Why no dropdown? Without it, certain window sizes get funky.

image

droumis: Also, without the use of the dropdown, the left side bar and top nav bar become quite redundant for this site.

droumis (commented):
This is definitely a big improvement! Apart from trying to use a dropdown to simplify the top nav, I support this being merged. 

The difference in top nav between holoviz.org and the blog website should continue to be addressed on the blog repo.

doc/conf.py
maximlt: I think the bug you noticed is a bug in the theme?

I set the dropdown menu to appear *Topics*:
image

PR #384, opened 2023-12-01
Update roadmap to clarify large data and workflows examples; add multi select groups and drilldown

droumis:
This PR updates the public roadmap to clarify/generalize the targets concerning large data handling and domain workflow examples. It also adds the internally known desiderata of multiple selection groups and drilldown support.

Besides a couple of other wording/whitespace fixes, I've also removed the solicitation for funding via a private entity which is noncompliant with NumFOCUS fiscal sponsorship.

I apologize for the obfuscation caused by the word wrap limit update. I noticed it too late.

### DIFF
--- a/doc/Roadmap.rst
+++ b/doc/Roadmap.rst
@@ -136,18 +136,10 @@ well with HoloViz include:
    Altair and Vega-lite visualizations can be specified and then used
    directly with Panel, but not with HoloViews.
 
-10. **hvPlot/HoloViews serialization**: HoloViews uses a declarative design that
-   can be represented in a purely textual form, without any Python code.
-   An initial implementation allows any Param-based objects (including
-   HoloViews objects) to be represented in JSON or YAML, but it needs
-   some polishing before it can be put into wide use for saving and
-   restoring configurations and layouts.
+11. **Multiple selection groups**: HoloViews currently supports a single
+   selection group. Multiple selection groups would allow for more flexible use
+   of, and comparison between, selections.
 
-11. **Better integration with ____**: There are a lot of tools in the
-   Python and other scientific software ecosystems that could be
-   included in HoloViz or made easily usable from it. NetworkX (already
-   usable but not fully exploited yet) is just one example of many;
-   suggestions welcome!
 
 If any of the functionality above is interesting to you (or you have
 ideas of your own!) and can offer help with implementation, please
--- a/doc/Roadmap.rst
+++ b/doc/Roadmap.rst
@@ -46,26 +46,16 @@ HoloViz include:
     that will help people get started and decide between the alternatives would
     be very helpful.
 
-2.  **Bokeh WebGL support**: Bokeh provides good support for working
-    interactively with small datasets using an HTML Canvas (client-side
-    interactivity), and when combined with Datashader it can handle enormous
-    datasets by pre-rendering them to images in Python (server-side). However, in
-    between these two extremes there is a range of data sizes that could be
-    addressed well by the client-side WebGL technology. Bokeh has long included
-    some WebGL support, but improvements in 2021-2022 have made the WebGL support
-    much more solid and usable, and updating HoloViews to use that could bring
-    big performance improvements for midsize plots.
-
-3.  **New and improved workflow examples in biomedical and Earth sciences,
+2.  **New and improved workflow examples in biomedical and Earth sciences,
     especially utilizing machine learning**: Demonstrations for handling and
     visualization of various large, multidimensional datasets from data-intensive
     research domains, including on HPC systems.
 
-4.  **Better deployment and exploration of large-scale visualizations**: For
+3.  **Better deployment and exploration of large-scale visualizations**: For
     instance, with improvements to automatic generation of tiles for exploration
     of large datasets using standard web servers.
 
-5.  **Integrating 3D surface and volume rendering into HoloViz**: HoloViews can
+4.  **Integrating 3D surface and volume rendering into HoloViz**: HoloViews can
     plot limited quantities of 3D surface or point data using `Matplotlib
     <http://holoviews.org/reference/elements/matplotlib/TriSurface.html>`__ or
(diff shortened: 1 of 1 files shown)


droumis:
Problem with formatting. It was there before this PR. UPDATE: fixed

image

doc/Roadmap.rst
```diff
@@ -1,158 +1,156 @@
...
+    available interactively in Bokeh, and similar support can be added to other
+    libraries by their users or contributors.
+
+3.  **Support for managing annotations**: Users of a plot or dashboard often
+    notice important features of the data on display, but don't have a convenient
+    way to capture or communicate that insight. The core team is developing
+    annotations as a first-class component the HoloViz tools, easily added onto
+    plots and dashboards so that users can denote regions of interest in some
+    space (e.g. a time range on a timeseries plot), storing, restoring, and
+    displaying that information in editable form.
```
jbednar:
```suggestion
    displaying that information in editable form. See https://github.com/holoviz/holonote for the work as it develops.
```

doc/Roadmap.rst
```diff
@@ -1,158 +1,156 @@
...
+
+Other targets needed in HoloViz or in packages designed to work well with
```
jbednar:
```suggestion
Other goals for HoloViz or packages designed to work well with
```

doc/Roadmap.rst
```diff
@@ -1,158 +1,156 @@
...
+
+2.  **Bokeh WebGL support**: Bokeh provides good support for working
+    interactively with small datasets using an HTML Canvas (client-side
+    interactivity), and when combined with Datashader it can handle enormous
+    datasets by pre-rendering them to images in Python (server-side). However, in
+    between these two extremes there is a range of data sizes that could be
+    addressed well by the client-side WebGL technology. Bokeh has long included
+    some WebGL support, but improvements in 2021-2022 have made the WebGL support
+    much more solid and usable, and updating HoloViews to use that could bring
+    big performance improvements for midsize plots.
```
jbednar: @Hoxbro , can you confirm -- can we now delete this WebGL entry? I'm pretty sure this has been completed and WebGL is now the default for HoloViews.

doc/Roadmap.rst
```diff
@@ -1,158 +1,156 @@
...
+
+3.  **New and improved workflow examples in biomedical and Earth science,
```
jbednar:
```suggestion
3.  **New and improved workflow examples in biomedical and Earth sciences,
```

(to me the plural reads better, since biomedical science is a separate field from Earth science, but it's just a preference).

doc/Roadmap.rst
```diff
@@ -1,158 +1,156 @@
...
+    especially utilizing machine learning**: Demonstrations for handling and
+    visualization of various large, multidimensional datasets from data-intensive
+    research domains, including on HPC systems.
+
+4.  **Better deployment and exploration of large-scale visualizations**: For
+    instance, with improvements to automatic generation of tiles for exploration
+    of large datasets using standard web servers.
```
jbednar: Maybe link to the last version of those examples on Datashader that got deleted, to point people to what to pick up?

doc/Roadmap.rst
```diff
@@ -1,158 +1,156 @@
...
+10. **hvPlot/HoloViews serialization**: HoloViews uses a declarative design
+    that can be represented in a purely textual form, without any Python code. An
+    initial implementation allows any Param-based objects (including HoloViews
+    objects) to be represented in JSON or YAML, but it needs some polishing
+    before it can be put into wide use for saving and restoring configurations
+    and layouts.
+
+11. **Multiple selection groups**: HoloViews currently supports a single
+    selection group. Multiple selection groups would allow for more flexible use
+    of, and comparison between, selections.
```
jbednar: I think this has to be explained better, as I don't think "selection group" is a common term in our docs.

doc/Roadmap.rst
```diff
@@ -1,158 +1,156 @@
...
+
+12. **Intuitive drilldown support**: In a drilldown action, the user
+    interacts in some way with one main plot (usually some low-dimensional
```
jbednar:
```suggestion
12. **Easy to configure drilldown support**: In a drilldown action, the user
    interacts in some way with one main plot (usually some low-dimensional
```

jbednar (approved):
Looks good! If you're reformatting the text anyway, please go all the way and simply remove the word wrapping, which should behave better for future diffs.

doc/Roadmap.rst
droumis: this? https://datashader.org/tiling.html

jbednar: The better link would be https://github.com/holoviz/datashader/blob/main/examples/tiling.ipynb , but for some reason I thought it had been deleted recently.

doc/Roadmap.rst
hoxbro: WebGL is now the default in HoloViews.

doc/Roadmap.rst
maximlt: Yes we currently discussed this notebook (and discovered it personally) as we were listing the S3 buckets we use and it's pointing to one. Jim you said:

> Datashader's tiling support is in exactly the state you are observing: a proof of concept, untested, unlinked in the docs, and yet still potentially useful for someone making tiles and discoverable by someone really dedicated. If rioxarray has better support, great! We should then make an example of using rioxarray with Datashader to render the tiles. If there is no open issue about dealing with tiles, there definitely should be; Brendan indeed left it in a half-finished state. I wouldn't worry about the S3 bucket, as it would appear to be for dumping the output, not holding precious input files.  So I'm inclined to do no action here besides having a better issue open if there isn't one, as we don't currently have any funding for tile-related work.

droumis: link to nb added

doc/Roadmap.rst
droumis: text revised

doc/Roadmap.rst
droumis: removed

PR #387, opened 2024-01-03
Add Governance to the site

maximlt:
I thought it'd be a good idea to expose HoloViz's governance docs to the community.

image
image
image

### DIFF
--- a/doc/about.rst
+++ b/doc/about.rst
@@ -77,28 +77,3 @@ HoloViz libraries have been supported by a variety of different funding mechanis
    <div style="text-align:center;">
      <img width="300px" src="https://raw.githubusercontent.com/numfocus/templates/master/images/numfocus-logo.png" alt="NumFOCUS Logo" style="display:inline-block;">
    </div>
-
-
-
-Governance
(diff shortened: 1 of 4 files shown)


droumis:
There is currently a Governance section on the About page that links out.  I don't have a strong opinion about elevating Governance to its own nav tab, but the content should probably be merged so that it's in one place.

droumis (commented):
There is currently a Governance section on the About page that links out. I don't have a strong opinion about elevating Governance to its own nav tab, but the content should probably be merged so that it's in one place. I'm a little bit concerned about the effect that adding more top-level pages has on the navigability of the website. But not concerned enough yet to block this. We can always reorganize/nest things later.

maximlt:
Good catch, I didn't know there was a Governance section on the About page :)

> I'm a little bit concerned about the effect that adding more top-level pages has on the navigability of the website.

I'm not too concerned about that since they're all displayed in the side bar on the landing page. I hope one day the theme allows us to better configure how things get displayed in the nav bar.

maximlt:
Done in https://github.com/holoviz/holoviz/pull/425

PR #386, opened 2024-01-03
Add HEPs to the website

maximlt:
HEPs were not exposed in the website, I'm suggesting they should be and this PR accomplishes that.

image
image
image

PR #388, opened 2024-01-08
HEP 2: Release and deprecation policies

maximlt:
Warning: builds on https://github.com/holoviz/holoviz/pull/386

---

This PR adds `HEP 2: Release and deprecation policies`. It is motivated by a discussion we had in https://github.com/holoviz/panel/pull/5221, and by my interest in ensuring that the HoloViz projects apply consistent policies, both for its users and maintainers. The HEP formalizes practices that have been in place for a long time and adds a few more rules on top of those. For instance, I've tried to come up with a clearer process for maintainers to adopt when they deprecate a feature. I've also suggested a minimum deprecation period of 18 months before removal, I'm sure there will be opinions about this period, and I'm also quite convinced we need to define such a period.

I drew some inspiration from:
- https://docs.djangoproject.com/en/dev/internals/release-process/#release-cadence
- https://numpy.org/neps/nep-0023-backwards-compatibility.html

I haven't touched all things on this topic, either because I didn't want to or didn't know what to suggest, I'd say these can be out of scope for now:

- I've addressed deprecation + removal, the HEP however says nothing about what should be done when changing the behavior of some API without removal (e.g. announcing the change of a default parameter value)
- Can a single maintainer decide to move/remove some API without getting anyone else involved? What if maintainers/contributors/users disagree?


doc/heps/hep2.md
```diff
@@ -0,0 +1,62 @@
...
+- Documentation: all usage of the deprecated feature must be removed from the documentation, except from the section that serves as the reference API. Special cases (e.g. major API change) will need extra documentation.
+- Release notes: the release notes must list the deprecated feature.
+
+When a feature has been deprecated, its warning and documentation must stay in place for some time long enough to let most users find out about it. For example, Panel users who lock the dependencies of their application should be given sufficient time between the deprecation of a feature and its removal so as not to miss the deprecation warnings and be left with broken code, once they attempt to upgrade their version of Panel to the latest. A deprecated feature **can only be removed 18 months after users have been informed of its deprecation**. This period, introduced by this HEP, factors in the facts that:
+
+- Users cannot build expectations about when in time a feature is going to be removed if the deprecation indicates a version number (e.g. `'This feature will be removed in version 1.2.3'`) as the Projects don't release new versions based on a regular cadence.
```
hoxbro: I don't buy this. Adding a version when the removal is being made is fine for both the users and us as developers. 

I also think defining the deprecation period of 18 months is too long for some packages (Panel) because a lot of releases can have happened in that time period, whereas for other packages, it may not be enough (colorcet). In a worst-case scenario, a user could download a package on the last day of the 17th month and start using it, only to see it removed the next day because of a new release.  

By keeping a table, updating it, and checking it, we also add extra burden on us as maintainers to keep it up to date. Having the deprecations as close to the code as possible makes it easier to avoid it from getting out of sync.

Again, I'm very much on having at least the option to put the version where the removal is gonna happen in the warning message, like what [Matplotlib](https://github.com/matplotlib/matplotlib/blob/eb02b108ea181930ab37717c75e07ba792e01f1d/lib/matplotlib/_api/deprecation.py#L32) and [xarray](https://github.com/pydata/xarray/blob/b35f7617480f857ed30f7fa494da41c7a7196237/xarray/util/deprecation_helpers.py#L102-L105) does. To be clear, this is not what all packages do. An example of this is pandas, which postpone it to the future, but they seem to be moving towards more major releases with breaking changes.

doc/heps/hep2.md
```diff
@@ -0,0 +1,62 @@
...
+- `beta` (e.g. `1.2.3b1`): Projects can deliver beta releases for the same reasons, except they indicate the project is in a more advanced development phase of a new version. beta releases are not common across the Projects, they're more likely to be useful when a major release is in development, to progressively deliver large new features and API breaking changes.
+- `release candidate` (e.g. `1.2.3rc1`): Projects must deliver a release candidate version before the final release.
+
+## Supported versions
+
+Exception in certain exceptional cases, the Projects do not backport changes to previous major or minor releases.
+
+## Release process
+
+Releasing a new version consists of packaging the software and distributing it. The Projects are set up to automatically perform this process on a Continuous Integration (CI) system (e.g. Github Actions). The process is triggered automatically when a new tag (of the form `v1.2.3`) is pushed to the Git repository. The tag message should be of the form `Version x.x.x` for final releases and `Version x.x.x. alpha1/beta1/RC1` for pre-releases.
```
hoxbro: I think the tag message should be removed or at least loosened, as it is, in my opinion, not important.

doc/heps/hep2.md
```diff
@@ -0,0 +1,62 @@
...
+
+While this versioning scheme is inspired by [Semantic Versioning](https://semver.org/) (*SemVer*), like many other Python projects, **the HoloViz Projects do not strictly follow it** as incompatible (aka breaking) API changes are not limited to `major` releases.
+
+The Projects can deliver three types of pre-releases / development versions:
+
+- `alpha` (e.g. `1.2.3a1`): Projects can deliver alpha releases for users to benefit from and/or test bug fixes and new features, early in the development phase of a new version. alpha releases are common across the Projects.
+- `beta` (e.g. `1.2.3b1`): Projects can deliver beta releases for the same reasons, except they indicate the project is in a more advanced development phase of a new version. beta releases are not common across the Projects, they're more likely to be useful when a major release is in development, to progressively deliver large new features and API breaking changes.
+- `release candidate` (e.g. `1.2.3rc1`): Projects must deliver a release candidate version before the final release.
```
hoxbro: This could also explicitly say that no new features can be added after the first RC, and only regressions and bugs should be handled. 

Must is also a strong word. I would say another of the dev releases is also good enough for a patch release. 

An expectation of a release candidate for some is that it has a time period where people have a chance to test it. What I see in our cases is that we make a release candidate to test if our CI runs correctly (with downstream test), and when it is finished, we make the official release.

hoxbro (commented):
Overall, this HEP is great.

However, I'm very much against being unable to put a version of when a deprecation is removed. 

Some technical details have also made it into this HEP. An example of this is the release project section. It is not that it is wrong, but if we, at some point, want to change it to no longer be triggered by pushing a tag, we need to update this HEP and get the approval.

doc/heps/hep2.md
maximlt: > Adding a version when the removal is being made is fine for both the users and us as developers.

My main goal here is to define a minimum deprecation period that all HoloViz projects can adhere to. I think it'd have benefits for its users and maintainers. I could see allowing projects to specify the removal version **as long as they ensure that it meets the minimum deprecation period**. For example, if a project releases `1.0.0` and announces a feature will be removed in `1.3.0`, the feature can only be removed X months after `1.0.0`, even if `1.3.0` happens before.

> I also think defining the deprecation period of 18 months is too long for some packages (Panel) because a lot of releases can have happened in that period, whereas for other packages, it may not be enough (colorcet). 

I don't think the number or frequency of releases matters so much for the deprecation period. The deprecation period is mostly a tradeoff between letting users have enough time to adapt (apps and libraries) and reducing the cost of maintaining a project. I want HoloViz as a whole project to agree on a minimum deprecation period. Individual projects are free to adopt longer deprecation periods.

> In a worst-case scenario, a user could download a package on the last day of the 17th month and start using it, only to see it removed the next day because of a new release.

This isn't a very good counter-example, as this is all about probability, and in that case well, bad luck! The same could happen with a shorter or longer deprecation period. Just to be clear, I'm not saying you should make a new release right after the deprecation period is over.

> By keeping a table, updating it, and checking it, we also add extra burden on us as maintainers to keep it up to date. Having the deprecations as close to the code as possible makes it easier to avoid it from getting out of sync.

I don't think HoloViz projects are used to deprecating and removing features, while I'd argue they should do it more to clean up their ever-growing API that sometimes confuses their users. For instance, I personally rarely check whether some code could be removed when I do a new minor release. I haven't built that muscle working on HoloViz projects. How about other maintainers? So I'm suggesting here a systematic approach to dealing with deprecated features, I don't think it's so much extra burden, and I anyway think it's part of the maintainer's job and more precisely of the release manager.

> Again, I'm very much on having at least the option to put the version where the removal is gonna happen in the warning message, like what [Matplotlib](https://github.com/matplotlib/matplotlib/blob/eb02b108ea181930ab37717c75e07ba792e01f1d/lib/matplotlib/_api/deprecation.py#L32) and [xarray](https://github.com/pydata/xarray/blob/b35f7617480f857ed30f7fa494da41c7a7196237/xarray/util/deprecation_helpers.py#L102-L105) does. To be clear, this is not what all packages do. An example of this is pandas, which postpone it to the future, but they seem to be moving towards more major releases with breaking changes.

- Matplotlib has great docs on how to deprecate a feature (https://matplotlib.org/devdocs/devel/api_changes.html#rules), it's a minimum of 2 minor releases later and possibly more. Its 3.x releases were: 3.8 (09/2023), 3.7 (02/2023), 3.6 (06/2022), 3.5 (11/2021), 3.4 (03/2021), 3.3 (07/2020), 3.2 (03/2020), 3.1 (05/2019), 3.0 (09/2018). The time between two minor releases is in months: 18, 14, 12, 16, 15, 15 and 15. It's a pretty stable release cadence so I think it's fair for them to announce they will remove a feature in 2 releases. HoloViz projects could also decide to adopt a regular-ish release cadence, though it might be more extra burden than what I'm suggesting.
- Xarray's contributor docs doesn't say you should say when a feature is removed, maybe their docs and code isn't in sync https://docs.xarray.dev/en/stable/contributing.html#backwards-compatibility.

doc/heps/hep2.md
maximlt: I suggested that because I observed that the message was pretty inconsistent across releases in a project, and across projects. I like consistency. But you're also right that it is not the most important part of the HEP.

doc/heps/hep2.md
maximlt: > This could also explicitly say that no new features can be added after the first RC, and only regressions and bugs should be handled.

Yes, I didn't think about that this is interesting.

> Must is also a strong word. I would say another of the dev releases is also good enough for a patch release.

I'm suggesting replacing the alpha releases (often just one) made before a patch release with release candidates.

> An expectation of a release candidate for some is that it has a time period where people have a chance to test it.

Yes, that's true. The HEP could explain that release candidates are always announced sometime before the final release.

> What I see in our cases is that we make a release candidate to test if our CI runs correctly (with downstream test), and when it is finished, we make the official release.

Not always, we have applications that test release candidates before the final release. It's not the most common case though, even if it's difficult to say whether other users rely on dev releases in their internal code bases.

maximlt:
Thanks for your review @Hoxbro !

> However, I'm very much against being unable to put a version of when a deprecation is removed.

I've commented on that in https://github.com/holoviz/holoviz/pull/388#discussion_r1453310581.

> Some technical details have also made it into this HEP. An example of this is the release project section. It is not that it is wrong, but if we, at some point, want to change it to no longer be triggered by pushing a tag, we need to update this HEP and get the approval.

Yep true that's a bit too technical and I can remove it, it's already documented in the [HoloViz Contributing Guide](https://holoviz.org/contributing.html).

But I have to say I'm not sure yet what exactly could and should get into a HEP. For instance, Matplotlib developer docs (https://matplotlib.org/devdocs/devel/) include a section with MEPs and many other sections on e.g. [Dependency version policy](https://matplotlib.org/devdocs/devel/min_dep_policy.html) (the equivalent of our HEP 1) or [API Changes](https://matplotlib.org/devdocs/devel/api_changes.html) (the equivalent of this HEP).

So in which case should we need a new HEP (if we need HEPs at all?)? @droumis pinging you since you told me you were thinking about writing HEP0.

maximlt:
I just talked with Simon and told him I was open to adopting another way to define the deprecation period, as long as there's some sort of consistency across the projects. I suggested a minimum deprecation period of 18 months, he prefers a minimum number of minor versions (correct me if I'm wrong). I still prefer a minimum deprecation period, but I think that in the end whatever we adopt will be better than the status quo!

maximlt:
Simon opened https://github.com/holoviz/geoviews/pull/701 to remove some features of GeoViews. I started to have a look at it and felt it was a good use case, and well, it was just in front of my eyes at the right time! This helps me understand the whole deprecation process and will hopefully help improve the HEP. By the way, this is all very factual, I don't intend to blame anyone :)

### Shared

- The deprecations were released in GeoViews 1.10.0 on the 25th of May 2023. They were planned to be removed one minor release later in GeoViews 1.11. Version 1.11.0 was released on the 1st of November 2023, 5 months later. Removing the deprecations was not done in this first patch release, it seems this is going to happen in version 1.11.1.
- It looks like GeoViews has some code to error when the removal version matches the current working version (in https://github.com/holoviz/geoviews/blame/main/geoviews/_warnings.py#L63). I guess it didn't catch this because the deprecated features were not exerted in the test suite.
- The warning emitted is a `GeoviewsDeprecationWarning`, a subclass of `DeprecationWarning`. Given how users run their code, this type of warning is not always emitted (see the section at the end)

### `load_tiff`

- It's not clear to me why `gv.util.load_tiff` was deprecated in the first place, one user reported in https://github.com/holoviz/geoviews/issues/639 that this function was using a deprecated API. Maybe `load_tiff` could have been adapted instead.
- Reading the PR that deprecated `gv.util.load_tiff` (https://github.com/holoviz/geoviews/pull/640), it isn't clear to me that the class methods `Image.load_tiff` and `RGB.load_tiff` were also deprecated.
- The deprecation was included in the release notes (Github and website) of version 1.10.0. The version targeted for removal was not mentioned there (thought it could have been inferred from the indication that the Wikipedia tile source was to be removed in 1.11.0)
- The deprecation was not mentioned in the docstrings. GeoViews has no API docs.
- `gv.util.load_tiff` isn't used in the documentation, so it didn't have to be updated :)
- The programmatic warning included an alternative with `geoviews.util.from_xarray(rioxarray.open_rasterio(filename)`. The release notes/documentation included an alternative too but not as clear as the programmatic one, with *`rioxarray.open_rasterio` to load GeoTIFFs into a `xarray.DataArray`.*
- `load_tiff` isn't used much out there (I can't tell if it was more used before 1.10 was released). The first two code bases have no test suite.
  - https://github.com/venuswku/data-visualizer/blob/eebc31e71f499b4f37ba48861c8abe9c62d2bd5f/data_visualizer/components/DataMap.py#L403
  - https://github.com/GlacioHack/geoprofile/blob/0a48380f05bf2f8824d98ece14237f537e345f5c/geoprofile/core/core.py#L357
  - EarthSim:
    - https://github.com/holoviz-topics/EarthSim/blob/5d6bde55feaa6e95e292a90575dfc5179f5b3beb/earthsim/grabcut.py#L384
    - (`gv.RGB.load_tiff`) https://github.com/holoviz-topics/EarthSim/blob/5d6bde55feaa6e95e292a90575dfc5179f5b3beb/examples/user_guide/WMTS_Example.ipynb#L102

### `Wikipedia` WMTS tile

- The issue (https://github.com/holoviz/geoviews/issues/623) and the PR (https://github.com/holoviz/geoviews/pull/630) don't mention why this is deprecated.
- https://github.com/holoviz/holoviews/pull/4941 (from May 2021) has more context on this deprecation. I have no time to dig more into it, there's a lot of discussion about how things should be deprecated, with a Parameter controlling some of the warning/error behavior. It seems the way this was deprecated in HoloViews vs GeoViews was quite different. The deprecated API was removed in HoloViews in https://github.com/holoviz/holoviews/pull/5654 (March 2023). The default tile was changed from Wikipedia to OSM in hvPlot in https://github.com/holoviz/hvplot/pull/610 (May 2021), the documentation wasn't updated accordingly (I'll open a PR for that).
- The deprecation was included in the release notes (Github and website) of version 1.10.0, stating the version targeted for removal was 1.11.0.
- `Wikipedia` is an instance of `WMTS`, there wasn't any docstring to update :)
- The documentation was updated in the PR, replacing `Wikipedia` with `OSM`.
- The programmatic warning included an alternative with `OSM`. The release notes/documentation suggested users to pick `OSM` too.
- It seems `gv.tile_sources.Wikipedia` has some more usage, some examples:
  - https://github.com/umr-lops/xsar/blob/cd3bfa72afaae95aa544f2efeb8f5c7f29a3ad79/docs/examples/xsar_tops_slc.ipynb#L50
  - https://github.com/umr-lops/xsarsea/blob/b12c73140137ba95197f8693c648c0e2d52bea67/docs/examples/streaks.ipynb#L34
  - https://github.com/OpenEnergyPlatform/open-MaStR/blob/bfa686f509a19a447610d3073fc40cfe07737406/postprocessing/helpers.py#L30
  - https://github.com/otro-mas1/ZBS_Madrid/blob/1dcee97e045c576188f7f6b5a1bf80c8a6f1c62f/ZBS_Mapa_Interactivo.ipynb#L79
  - https://github.com/fccoelho/epigrass/blob/5a94cbc236dbfc60a82d71e8af4ab367779bf44c/Epigrass/epipanel.py#L11

---

I hope I don't get this wrong as Python warnings confuse me a little, but running the code below with `python foo.py` doesn't print the warning. It does print it if I disable the default warning filters with `python -W default foo.py`, showing:

> /home/liquetm/dev/bar.py:5: DeprecationWarning: Bottom is deprecated
  return bottom()

```py
# foo.py
from bar import top

top()
```

```py
# bar.py
import warnings

def top():
    return bottom()

def bottom():
    warnings.warn('Bottom is deprecated', DeprecationWarning, stacklevel=2)
```

maximlt:
My take away from this as a maintainer:

- I'd like us to reach a consensus on how to deprecate a feature, the difference between how the Wikipedia tile source was deprecated in HoloViews and GeoViews is quite big, this can only lead to endless discussions
- I'd like much better guidelines, the whole deprecation cycle involves many actions and is quite complex to get right

maximlt:
Made some changes:

- Removed the too detailed section on the release process as suggested by Simon (https://github.com/holoviz/holoviz/pull/388/commits/989e93535032c9b5cce0988db263a3577324048a)
- Clarified definition/expectation of release candidates (https://github.com/holoviz/holoviz/pull/388/commits/1e5a7ad0945d1dde0543f6f476643c6bc6346878)
- Improved the deprecation guidelines and added an explanation on Python warnings (https://github.com/holoviz/holoviz/pull/388/commits/44f52dd099ea03023251e92febe05c54eae07de3)

I think more work is needed:

- [x] (Re)-exploring the various ways Python warnings are filtered/displayed depending on their type and the context made me realize the guidelines should be more clear about them.
- [x] I suggested maintainers should update a deprecation table somewhere, I want to see how this might be done with Param. Simon on his side has looked into exploring a code base to find its warnings and parse them (with the AST). It'd be great to find a systematic approach that isn't too over-engineered.

maximlt:
I finally got around to dedicating some time to the HEP and updated it in https://github.com/holoviz/holoviz/pull/388/commits/43cb23991e1163fd0836f9ece9acfe7a526a4936. I probably need to give it a good review with fresh eyes, but the next step for me will be to introduce it to the team in a HoloViz meeting.

maximlt:
@philippjfr here are some deprecation periods for other libraries:

- [Django](https://docs.djangoproject.com/en/dev/internals/release-process): 
  - Feature releases are made roughly every 8 months. Deprecations need to wait for a minimum of 2 feature releases (carrying over major versions if needed). For example, Django 4.2 deprecated something, next version is Django 5.0, the feature is removed in Django 5.1.
  - So this is a minimum of 16 months
- [Numpy](https://numpy.org/neps/nep-0023-backwards-compatibility.html):
  - Removal shall be done after at least 2 releases assuming the current 6-monthly release cycle; if that changes, there shall be at least 1 year between deprecation and removal. Removal can be done in any minor, but not bugfix, release.
- [Matplotlib](https://matplotlib.org/devdocs/devel/api_changes.html#rules):
  - Removal minimum of 2 minor releases later and possibly more.
  - Matplotlib 3.x releases were: 3.8 (09/2023), 3.7 (02/2023), 3.6 (06/2022), 3.5 (11/2021), 3.4 (03/2021), 3.3 (07/2020), 3.2 (03/2020), 3.1 (05/2019), 3.0 (09/2018). The time between two minor releases is in months: 18, 14, 12, 16, 15, 15 and 15.
- [Scipy](https://docs.scipy.org/doc/scipy/dev/core-dev/index.html#deprecations):
  - Wait till at least 6 months after the release date of the release that introduced the DeprecationWarning before removing the functionality

18 months is for sure quite high. Listening to your suggestions on:

- a shorter grace period
- special-casing deprecations made before a major release, perhaps with an even shorter grace period

maximlt:
@jlstevens suggested it would be great if there was automation to enforce some of the requirements. I think that:

- Ideally, yes, but building and setting up automation over multiple repos and in the correct way would require some effort
- Proper review of PRs deprecating and removing features should already be sufficient to make sure this is done correctly
- The most important part, which is to decide to deprecate something and to implement it, cannot be automated.

doc/heps/hep2.md
```diff
@@ -0,0 +1,370 @@
...
+</table>
+
+## Summary
+
+This HEP describes the release and deprecation policies adopted by the HoloViz Projects. While two HEPs could have be written to describe these policies separately, it is deemed that they are enough intertwined to be addressed together (e.g. to answer questions such as in which kind of final release can a feature be deprecated/removed?).
+
+The Projects already follow a pretty consistent procedure when it comes to releases, which this HEP formalizes. However, the Projects have applied various approaches for their deprecation cycle, and without clear guidelines this has sometimes led to long and unproductive discussions. The HEP aims to improve this by introducing new and well defined guidelines. Making it easier for Projects to deprecate features is **key to their long-term maintenance**.
+
+The overall goal of the HEP is to ensure that these policies are **consistently applied across the Projects**, which will help provide a **consistent user experience** and **help contributors and maintainers make decisions**. We also aim to adopt **standard practices**.
+
```
droumis: Let's add a concise 'Resolution' section after 'Summary' for a quick recap of the technical plan of action. Something like:

```
## Resolution
- Adopt a tiered warning strategy for deprecations:
  - Initially emit a `DeprecationWarning`, ensuring it's set with the appropriate stacklevel for accurate source identification.
  - After a 12-month period, elevate the warning to a `FutureWarning` to capture the attention of all users, including those who might not be configured to see DeprecationWarnings.
- Allow exceptions in rare cases, such as beginning with a `PendingDeprecationWarning` for potentially reversible or highly disruptive changes, or immediately using a `FutureWarning` for urgent cases.
```

doc/heps/hep2.md
```diff
@@ -0,0 +1,370 @@
...
+<tr><td> Author(s) </td><td> Maxime Liquet </td></tr>
+<tr><td> Created </td><td> TBD </td></tr>
+<tr><td> Updated </td><td> - </td></tr>
+<tr><td> Discussion </td><td> https://github.com/holoviz/holoviz/pull/388 </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+## Summary
+
+This HEP describes the release and deprecation policies adopted by the HoloViz Projects. While two HEPs could have be written to describe these policies separately, it is deemed that they are enough intertwined to be addressed together (e.g. to answer questions such as in which kind of final release can a feature be deprecated/removed?).
```
droumis: I think the 'Summary' and 'Motivation' are a bit too overlapping. I'd suggest simplifying the Summary such as follows:

"This HEP outlines unified release and deprecation policies for the HoloViz Projects. It  addresses key questions about the lifecycle of features, including deprecation and removal timelines. The HEP formalizes existing release procedures and introduces well-defined deprecation guidelines to streamline long-term maintenance and decision-making processes, aiming for consistency across projects and an improved user experience."

droumis (commented):
Looks good, added a couple suggestions

doc/heps/hep2.md
maximlt: I don't really want to summarize the plan of action as I think all the steps of the deprecation cycle are important.

jlstevens:
If automation is too high a bar, then the next best thing is establishing a clear, concrete process that does things correctly. For instance, maybe we should have a standardized approach/template for a 'deprecation PR' that we need to create every release. 

That way a maintainer can just look up at a reference deprecation PR (e.g. the previous one), follow the steps and without too much cognitive effort be fairly confident that the recommendations in this HEP have been followed. At a minimum I would want a standardized PR name (e.g. Deprecations for version X.X.X) and the first comment to link to this HEP!

doc/heps/hep2.md
maximlt: Thanks for the suggestion! Updated in https://github.com/holoviz/holoviz/pull/388/commits/7d42f6f99afc0dc18f3b42724c0a457f8e8d327b.

maximlt:
> If automation is too high a bar, then the next best thing is establishing a clear, concrete process that does things correctly. For instance, maybe we should have a standardized approach/template for a 'deprecation PR' that we need to create every release.
> 
> That way a maintainer can just look up at a reference deprecation PR (e.g. the previous one), follow the steps and without too much cognitive effort be fairly confident that the recommendations in this HEP have been followed. At a minimum I would want a standardized PR name (e.g. Deprecations for version X.X.X) and the first comment to link to this HEP!

But isn't the clear process the `Specification` section of the HEP?

jlstevens:
> But isn't the clear process the Specification section of the HEP?

Yes but you also want something concrete as well that integrates well into your release cycle. Checking the `Specification` section of all HEPs at release for instructions is a rather painful workflow.

I'm trying to suggest an approach to make our lives easier, not harder! :-)

doc/heps/hep2.md
```diff
@@ -0,0 +1,366 @@
...
+
+### Release policy
+
+#### Versioning
+
+The Projects deliver final releases with a version number of the form `<major>.<minor>.<patch>` (e.g. `1.2.3`):
+
+- `patch` releases (e.g. 0.1.**0** -> 0.1.**1**) should never intentionally and rarely, if ever, unintentionally break API. Should be safe to upgrade to the latest patch release if you encounter any problems.
+- `minor` releases (e.g. 0.**1**.1 -> 0.**2**.0) should not have API changes that affect most users. API changes in minor releases should be rare, but are not unheard of, particularly for recently added functionality whose API is still being refined, or for bugs found in older code that can't be fixed without changing API at least slightly.
+- `major` releases (e.g. **0**.2.0 -> **1**.0.0)  will typically break APIs, but should normally include significant new functionality to motivate users to update their code. Major breaking API changes should be postponed to a major release and paired with significant user-visible advantages.
```
philippjfr: I would remove the sentence about "significant new functionality", I think we should be perfectly comfortable making a major release solely to make API changes.

doc/heps/hep2.md
```diff
@@ -0,0 +1,366 @@
...
+
+While this versioning scheme is inspired by [Semantic Versioning](https://semver.org/) (*SemVer*), like many other Python projects, **the HoloViz Projects do not strictly follow it** as incompatible (aka breaking) API changes are not limited to `major` releases.
+
+The Projects can deliver three types of pre-releases / development versions:
```
philippjfr: Does this belong here? I don't think it's strictly related to the deprecation discussion.

philippjfr: Ah, sorry, I realize this HEP is meant to cover both release policies and deprecation. Personally I'd almost say those are two separate proposals.

doc/heps/hep2.md
```diff
@@ -0,0 +1,366 @@
...
+
+The HoloViz Projects serve different purposes in the ecosystem. Some are more foundational and have been in place for a long time, for instance, Colorcet or Param. Given their position, these Projects should be treated with extra care when it comes to backwards compatibility. For instance, maintainers of these Projects should favor making breaking changes in major releases and adopting longer deprecation periods. On the other hand, some Projects are newer and have an API that is still being gradually refined (e.g. Lumen). These Projects are not expected to be as stable, they change more quickly. 
+
+Overall, the HoloViz Projects are known to be stable and their users have built this expectation. The functionalities they provide are generally not moved or removed lightly. Maintainers should aim to keep the Projects stable; moving or removing a feature from a code base **must** be motivated, in particular when it is done outside of a major release.
+
+### Deprecation policy
+
+#### Deprecation cycle guidelines
+
+The Projects can deprecate a feature in any type of final release (patch, minor, or major). However, before implementing the deprecation, they **must**:
```
philippjfr: I would say new deprecations should never be issued in a patch release.

doc/heps/hep2.md
```diff
@@ -0,0 +1,366 @@
...
+
+- Motivate the best they can why the feature is deprecated (e.g. old and unused API) in an issue or in the *Pull Request* (PR) deprecating the feature.
+- Reach a consensus among the maintainers of the Project on the deprecation.
+- Make sure that the PR deprecating the feature is approved by at least another maintainer, who should check that the indications below have been followed.
+
+ Maintainers have to make sure that their users are well informed of the deprecation. They **must** mention the deprecation in the following ways:
+
+- Implement a programmatic warning, if not applicable the PR deprecating the feature **must** clearly explain why:
```
philippjfr: Should we not standardize on one mechanism or are we worried about the `typing_extensions` dependency?

doc/heps/hep2.md
```diff
@@ -0,0 +1,366 @@
...
+- Documentation: all usage of the deprecated feature **must** be removed from the documentation, except from the section that serves as the reference API. Special cases (e.g. major API change) will need extra documentation.
+- Release notes: the release notes **must** list the deprecated feature.
+- Repository: all the open issues an PRs associated with the deprecated feature **must** be closed.
+
+When the deprecation involves the user having to change their code in a significant way (typically in a major release), maintainers should consider writing a migration guide.
+
+When a feature has been deprecated, its warning and documentation **must** stay in place for some time long enough to let most users find out about it. For example, Panel users who lock the dependencies of their application should be given sufficient time between the deprecation of a feature and its removal so as not to miss the deprecation warnings and be left with broken code, once they attempt to upgrade their version of Panel to the latest. A deprecated feature **can only be removed 18 months after users have been informed of its deprecation**. This period, introduced by this HEP, factors in the facts that:
```
philippjfr: I don't think prescribing a fixed time period really works for us. In theory semantic versioning solves this problem by setting clear expectation that a major release can break things at will, even without any announcement whatsoever. If people care about that they can appropriately pin major releases.

My main point here is that we can attempt to provide some predictability around the API with this but in practice there are major changes that can happen with relatively little lead time, e.g. think of the Bokeh layout refactor and its effect on Panel 1.x. Until about a month or two before the release it was still quite unclear what the actual effect of the change would be, so despite the fact that API changes are telegraphed that actually guarantees very little about compatibility between major versions.

doc/heps/hep2.md
```diff
@@ -0,0 +1,366 @@
...
+| `PendingDeprecationWarning` | Base category for warnings about features that will be deprecated in the future (ignored by default).                                                                         |
+| `ImportWarning`             | Base category for warnings triggered during the process of importing a module (ignored by default).                                                                           |
+| `UnicodeWarning`            | Base category for warnings related to Unicode.                                                                                                                                |
+| `BytesWarning`              | Base category for warnings related to bytes and bytearray.                                                                                                                    |
+| `ResourceWarning`           | Base category for warnings related to resource usage (ignored by default).                                                                                                    |
```
philippjfr: Would remove irrelevant warnings from this list.

doc/heps/hep2.md
hoxbro: Sometimes, a deprecation can be needed if something changes outside our control.

doc/heps/hep2.md
maximlt: The sentence says `should normally include significant new functionality` so it doesn't make it a strong requirement.

By the way, I literally copied these definitions from a comment from @jbednar https://github.com/holoviz/panel/pull/5221#issuecomment-1644901836.

doc/heps/hep2.md
maximlt: Yes. They're together because it's a bit difficult to talk about deprecations without having defined a clear version policy. So I think that if two separate proposals were made Version Policy should come first.

doc/heps/hep2.md
maximlt: > I would say new deprecations should never be issued in a patch release.

I don't have strong opinions on this, so I could add that to the HEP.

> Sometimes, a deprecation can be needed if something changes outside our control.

True, but we control what type of release we decide to make. And, there will always be exceptions to the rule.

doc/heps/hep2.md
maximlt: > I don't think prescribing a fixed time period really works for us. 

The HEP doesn't prescribe a fixed time period, instead, it sets a minimum time period.

> In theory semantic versioning solves this problem by setting clear expectation that a major release can break things at will, even without any announcement whatsoever. If people care about that they can appropriately pin major releases.

It doesn't work quite like that in practice right 🙃 

> My main point here is that we can attempt to provide some predictability around the API with this but in practice there are major changes that can happen with relatively little lead time, e.g. think of the Bokeh layout refactor and its effect on Panel 1.x. Until about a month or two before the release it was still quite unclear what the actual effect of the change would be, so despite the fact that API changes are telegraphed that actually guarantees very little about compatibility between major versions.

Yes, my main point is to add consistency and indeed predictability. I agree that major versions are a special case and I can update the HEP accordingly, but I would still recommend giving people enough time when possible. And yes, there will always be exceptions, but I hope there will be just a few of them. The example you give is typically the sort of thing that I hope would not have happened in such a short time, for example, that didn't leave so much time to plan how to migrate CSS.

doc/heps/hep2.md
maximlt: Sure I can do that.

doc/heps/hep2.md
maximlt: Given `deprecated` is a decorator I don't think is can be used in all the places you may want to use `warn`?

doc/heps/hep2.md
maximlt: Done in https://github.com/holoviz/holoviz/pull/388/commits/a5b5894d0bed29ee613bf26fbafadaf982500ba6

maximlt:
In the last round of updates I have:

- Attempted to make it clear that exceptions to the rules are fine
- Turned the deprecation cycle guidelines into a checklist (cc @jlstevens)
- Changed the "must respect a minimum period of 18 months" to "recommend a minimum period of 6 months", given the low interest I saw in adopting a strict minimum deprecation period, and especially one of 18 months. Yet, I still would like to try to ensure some consistency across the projects.

maximlt:
Another update to align some recommendations with the previous update:
- Upgrade from `DeprecationWarning` to `FutureWarning` after a few months
- When removing the feature, ensure the warning type was visible enough

maximlt:
@holoviz/holoviz-dev this is ready for another round of review. Let me know if I have not addressed one of your previous comments and if you disagree with how I addressed them.

maximlt:
Add two small things after discussing with Simon:

- Projects should distribute the release candidates before a major release and all final releases as *Github releases*
- The deprecation policy applies to what each Project considers to be its *user-facing public API*

doc/heps/hep2.md
```diff
@@ -0,0 +1,366 @@
...
+      1. Start by emitting a `DeprecationWarning`, to inform Library Developers, and some Data Analysts.
+      2. After a few months, upgrade to a `FutureWarning`, to inform all users.
+
+      Exceptions are allowed but should in practice be very rare:
+
+      - Start by emitting a `PendingDeprecationWarning` for deprecations that might be reverted or are expected to be extremely noisy.
+      - Don't upgrade to a `FutureWarning` when the feature removed is meant for Library Developers only.
+      - Emit directly a `FutureWarning` when the feature removed is meant for Data Analysts only.
+
+  - `stacklevel` **must** be set so the warning appears in the correct place. Projects can implement a utiliy like the private Pandas' function [`find_stack_level`](https://github.com/pandas-dev/pandas/blob/b8a4691647a8850d681409c5dd35a12726cd94a1/pandas/util/_exceptions.py#L34) to automatically infer the right `stacklevel` number.
```
ahuang11: I don't see stacklevel being used in `@deprecated`

doc/heps/hep2.md
```diff
@@ -0,0 +1,366 @@
...
+- [ ] Documentation: all usage of the deprecated feature **must** be removed from the documentation, except from the section that serves as the reference API. Special cases (e.g. major API change) will need extra documentation. When the deprecation involves the user having to change their code in a significant way (typically in a major release), maintainers should consider writing a migration guide.
+
+3. When the deprecation is fully implemented, maintainers:
+
+- [ ] **Must** make sure that the PR deprecating the feature is approved by at least another maintainer, who should check that the approach above has been followed.
+- [ ] **Must** close all the open issues an PRs associated with the deprecated feature.
+
+4. When releasing the deprecation, maintainers:
+
+- [ ] **Must** include it in a major of minor release, not in a patch release.
```
ahuang11: I think typo
```suggestion
- [ ] **Must** include it in a major or minor release, not in a patch release.
```

maximlt: Yes, thanks!

doc/heps/hep2.md
maximlt: Yep true, although it's part of its signature so there might be some reasons to use it for weird edge cases (can't think of one). I can update the text to reflect that. I also found out the runtime behavior of `@deprecated` can be disabled by setting `category` to `None` which means you can used both `@deprecated` and `warnings.warn` if needed. Which I think means that `typing_extensions` doesn't have to be added as a runtime dependency.

I also saw that `warnings.warn` gained the new `skip_file_prefixes` parameter in Python 3.12. If I understand what it does I could mention it in the HEP.

maximlt: Pushed some updates in https://github.com/holoviz/holoviz/pull/388/commits/b5def36563835f171f95306b9eb9ee12384d0303 and https://github.com/holoviz/holoviz/pull/388/commits/f8e00677152545d4cc01938fd371b19fa7792f1e about the above.

maximlt:
- https://github.com/holoviz/holoviz/commit/b5def36563835f171f95306b9eb9ee12384d0303: Enhanced the `warnings.warn` / `@deprecated` section by putting `@deprecated` first and documenting that calling `@deprecated(msg, category=None)` **disables** the run-time warning.
- https://github.com/holoviz/holoviz/commit/f8e00677152545d4cc01938fd371b19fa7792f1e: Added docs on the `skip_file_prefixes` parameter of `warnings.warn` added in Python 3.12 that I think could be used in place of setting `stacklevel` manually or with `find_stack_level`.

doc/heps/hep2.md
jbednar: Right after that comment I wrote "I'd be interested in whether other team members think this description is accurate.", but no one engaged with it.  Reading it now, I'm happy for @philippjfr to make the call -- would we _normally_ provide some carrot along with the stick, i.e. provide some enticement to make it worth switching? You make more releases than I do, so I leave it to your discretion.

doc/heps/hep2.md
jbednar: I think https://jacobtomlinson.dev/effver/ captures more of our process than SemVer does; maybe link to that.

doc/heps/hep2.md
maximlt: FWIW I re-used your description as I found it appropriate.

maximlt:
HEP2 is ready for review. I think the first reviews and discussions have helped resolve some points of contention, but there may still be some. According to HEP0, we should aim to reach a consensus. To make sure I capture everyone's perspective, please add a comment. It could be as simple as 👍, if there are parts of the HEP you disagree with please raise them so they can be further discussed.

cc @jbednar @philippjfr @jlstevens @MarcSkovMadsen @Hoxbro @droumis @ahuang11

doc/heps/hep2.md
```diff
@@ -0,0 +1,370 @@
...
+      - Emit directly a `FutureWarning` when the feature removed is meant for Data Analysts only.
+
+  - `stacklevel`, or `skip_file_prefixes` starting from Python 3.12, **must** be set when calling `warnings.warn` so the warning message references the right deprecated code line. Projects can implement a utility like the private Pandas' function [`find_stack_level`](https://github.com/pandas-dev/pandas/blob/b8a4691647a8850d681409c5dd35a12726cd94a1/pandas/util/_exceptions.py#L34) to automatically infer the right `stacklevel` number. Python 3.12 added the `skip_file_prefixes` to `warnings.warn` that could be used in place of setting `stacklevel` manually or with `find_stack_level`, with e.g. `skip_file_prefixes=os.path.dirname(importlib.util.find_spec(<package_name>).origin)`.
+
+- [ ] Documentation: all usage of the deprecated feature **must** be removed from the documentation, except from the section that serves as the reference API. Special cases (e.g. major API change) will need extra documentation. When the deprecation involves the user having to change their code in a significant way (typically in a major release), maintainers should consider writing a migration guide.
```
droumis: To me, it's a bit unclear what is meant by `except from the section that serves as the reference API`. Shouldn't we also put a message in the reference API that it is going to be deprecated? This would be helpful if/when we move to versioned docs.

droumis:
minor comment about docs, but either way 👍

droumis (approved):

doc/heps/hep2.md
maximlt: Great catch thanks! This was in the original commit and got lost over edits. I'll re-add this bit:

> Docstring (if applicable): the docstring of the deprecated feature must be updated to mention the deprecation, when possible the [deprecated](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated) Sphinx directive should be used.

maximlt: Done in https://github.com/holoviz/holoviz/pull/388/commits/1712fa4353e1be7e1fbf71bf7bcf90b02847bec2

doc/heps/hep2.md
```diff
@@ -0,0 +1,372 @@
...
+default::DeprecationWarning:__main__
+ignore::DeprecationWarning
+ignore::PendingDeprecationWarning
+ignore::ImportWarning
+ignore::ResourceWarning
+```
+
+Note that the default filters can be overriden by setting the `-W` flag when calling the `python` executable or via the `PYTHONWARNINGS` environment variable.
+
+The default filters imply that in a regular context `PendingDeprecationWarning` will not be seen by the user, while `FutureWarning` will always be seen. `DeprecationWarning` is treated specially; warnings of these types are only seen when they are triggered by and points to code in `__main__`, the latter depending on the value of `stacklevel`.
```
jbednar: I can't parse "seen when they are triggered by and points to code in".

doc/heps/hep2.md
```diff
@@ -0,0 +1,372 @@
...
+
+- they serve a Panel app with `panel serve`
+- or, run a script/Notebook that imports a utility module/package that calls a deprecated feature emitting `DeprecationWarning`
+
+### Outcome
+
+Based on the fact that a non-negligeable fraction of HoloViz users may entirely miss `DeprecationWarning`s, we suggest adopting a tiered approach:
+
+1. Start by emitting a `DeprecationWarning`, to inform Library Developers, and some Data Analysts.
+2. After 12 months, upgrade to a `FutureWarning`, to inform all users.
```
jbednar: Saying 12 months here seems to conflict with an overall deprecation time of 6 months mentioned elsewhere.

jbednar (approved):
I've given it a detailed review, and made a bunch of tiny wording fixes and clarifications, none of which I think should be controversial. I have a couple of minor questions below, but overall, I'm happy with it and happy for it to be merged. Good job! (And I vote for a 6 month period; longer than that and we will have no memory of what we were doing!)

doc/heps/hep2.md
```diff
@@ -54,7 +54,7 @@ Aside from certain exceptional cases, the Projects are not expected to backport
 HoloViz release managers are responsible for distributing the Projects on these platforms:
 
 - Pre-releases are distributed on [PyPI](https://pypi.org) and on the *pyviz/label/dev* channel of [Anaconda.org](https://anaconda.org).
-- Final versions are distributed on [PyPI](https://pypi.org), and on the  *conda-forge* and *pyviz* channels of [Anaconda.org](https://anaconda.org).
+- Final versions are distributed on [PyPI](https://pypi.org), and on the  *conda-forge* and *pyviz* channels of [Anaconda.org](https://anaconda.org), and (in many cases) on the *main* channel as well.
```
maximlt: @jbednar I didn't include the *main* channel originally as that feels like this shouldn't be the responsibility of a HoloViz maintainer, but of an Anaconda employee, and it turns out the intersection of both is close to 100% at the moment (it may not always be the case).

doc/heps/hep2.md
maximlt: Thanks I missed updating the appendix, fixed in https://github.com/holoviz/holoviz/pull/388/commits/6c7a533c8b0f60e77bec612a9d8cc1b1f8842649.

doc/heps/hep2.md
maximlt: Rephrased in https://github.com/holoviz/holoviz/pull/388/commits/c91e568e67be78b7025572e7dc503f9fbd851244

> `DeprecationWarning` is treated specially; warnings of these types are displayed only if `warn` is called by code located in the `__main__` namespace and if `warn` is configured to emit the warning in the `__main__` namespace, the latter depending on the value of `stacklevel`.

doc/heps/hep2.md
```diff
@@ -0,0 +1,372 @@
...
+- `beta` (e.g. `1.2.3b1`): Projects can deliver beta releases for the same reasons, except they indicate the project is in a more advanced development phase of a new version. beta releases are not common across the Projects. They are more likely to be useful when a major release is in development, to progressively deliver large new features and API breaking changes.
+- `release candidate` (e.g. `1.2.3rc1`): Projects must deliver at least one release candidate version before the final release and Projects should not add new features between release candidates. There is no expectation from Projects to announce their release candidates to a wide audience or to wait some time before making a final release. For major or significant minor releases, getting wider feedback in this way before release is encouraged.
+
+#### Supported versions
+
+Aside from certain exceptional cases, the Projects are not expected to backport changes to previous major or minor releases.
```
philippjfr:
```suggestion
Aside from certain exceptional cases (such as major security vulnerabilities), the Projects are not expected to backport changes to previous major or minor releases.
```

doc/heps/hep2.md
```diff
@@ -0,0 +1,372 @@
...
+
+#### Distribution
+
+HoloViz release managers are responsible for distributing the Projects on these platforms:
+
+- Pre-releases are distributed on [PyPI](https://pypi.org) and on the *pyviz/label/dev* channel of [Anaconda.org](https://anaconda.org).
```
philippjfr: Kind of unrelated but we should move to a `holoviz` channel eventually. Would be nice if we could set up an alias.

philippjfr (approved):
Left some minor comments but I'm quite happy with this! Really appreciate the effort you put into this.

doc/heps/hep2.md
maximlt: Yes, I'd love to move away entirely from `pyviz`. We've made good progress but we're not yet there :)

doc/heps/hep2.md
```diff
@@ -0,0 +1,372 @@
...
+    - **Must** suggest replacement APIs, if applicable.
+    - Can indicate in which version the feature was deprecated.
+    - Can indicate before which version the feature is going to be removed. However, maintainers should ensure that the deprecation period (see below) is not going to be too short, and that no obsolete warning will be released (e.g. feature announced to be removed in version 1.1 but is still present in 1.1).
+
+  - The warning type:
+  
+    - **Must** be a subclass of `DeprecationWarning`, `FutureWarning`, or `PendingDeprecationWarning`.
+    - **Must** be defined based on this approach (see the Appendix for more details):
+
+      1. Start by emitting a `DeprecationWarning`, to inform Library Developers, and some Data Analysts.
```
hoxbro:
```suggestion
      1. Start by emitting a `DeprecationWarning`, to inform Library Developers, and some users.
```

doc/heps/hep2.md
```diff
@@ -0,0 +1,372 @@
...
+
+#### Release cadence
+
+The Projects have not adopted a regular release cadence; a new release happens whenever maintainers find it is appropriate. A Project is free to adopt a regular release cadence.
+
+Major regressions in a release should be fixed and released in a new patch version as soon as possible.
+
+### Backwards compatibility
+
+The HoloViz Projects serve different purposes in the ecosystem. Some are more foundational and have been in place for a long time, for instance, Colorcet or Param. Given their position, these Projects should be treated with extra care when it comes to backwards compatibility. For instance, maintainers of these Projects should favor making breaking changes in major releases and adopting longer deprecation periods. On the other hand, some Projects are newer and have an API that is still being gradually refined (e.g. Lumen). These Projects are not expected to be as stable, and will change much more quickly.
```
hoxbro:
```suggestion
The HoloViz Projects serve different purposes in the ecosystem. Some are more foundational and have been in place for a long time, for instance, [Colorcet](https://github.com/holoviz/colorcet) or [Param](https://github.com/holoviz/param). Given their position, these Projects should be treated with extra care when it comes to backwards compatibility. For instance, maintainers of these Projects should favor making breaking changes in major releases and adopting longer deprecation periods. On the other hand, some Projects are newer and have an API that is still being gradually refined (e.g. [Lumen](https://github.com/holoviz/lumen)). These Projects are not expected to be as stable, and will change much more quickly. 
```

Just for convenience, and only for the first time it is mentioned.

doc/heps/hep2.md
```diff
@@ -0,0 +1,372 @@
...
+
+2. Maintainers have to make sure that their users are well informed of the deprecation. They **must** implement the deprecation in the following way:
+
+- [ ] Implement a programmatic warning; if not applicable the PR deprecating the feature **must** clearly explain why.
+
+  - The warning **must** be emitted using one or a combination of these two utilities:
+
+    - The `deprecated` decorator added in Python 3.13 ([PEP 702](https://peps.python.org/pep-0702/)) to the `warnings` module, and backported to previous Python versions via the [typing_extensions](https://github.com/python/typing_extensions) package, can be used to mark *functions*, *classes*, and *overloads* as deprecated. This decoration has two features:
+
+      1. It enables *static type* checkers to warn when they encounter usage of an object decorated with `@deprecated`. This comes with various benefits, for example, VSCode users will see the deprecated objects crossed out in their code.
```
hoxbro:
```suggestion
      1. It enables *static type* checkers to warn when they encounter usage of an object decorated with `@deprecated`. This comes with various benefits, for example, [LSP](https://en.wikipedia.org/wiki/Language_Server_Protocol) (like pylance in VSCode) users will see the deprecated objects crossed out in their code.
```

I think this is just an LSP feature, not a VSCode feature.

doc/heps/hep2.md
```diff
@@ -0,0 +1,372 @@
...
+
+Pytest automatically catches all warnings during test execution and displays them at the end of the session. The issues listed above with `DeprecationWarning`s displayed only in a certain scope don't apply in this case; these warnings are always displayed by Pytest.
+
+
+#### Two types of HoloViz users
+
+We can define two types of HoloViz users:
+
+- Library Developers: depend on HoloViz projects to develop their own library (this includes HoloViz projects themselves as they depend on each other, like Panel with Param)
+- Data Analysts: use HoloViz projects to perform some sort of data analysis, often by writing code in a Notebook or in scripts they run directly or by building and serving a Panel app.
```
hoxbro: I think we should just call this users.

hoxbro (approved):

doc/heps/hep2.md
maximlt: To me library devs are also *users* so I wouldn't know how to differentiate both. Not sure what other term would fit.

doc/heps/hep2.md
maximlt: EffVer is mentioned in the HEP.

doc/heps/hep2.md
hoxbro: You can be both. It is just different hats you take on. 

I see myself both as a user and developer of the HoloViz stack.

maximlt: Maybe there's a confusion. `Library Developers` means people using one or more HoloViews projects to develop their own library, not HoloViz developers like us.

hoxbro: I understand that. I just think data analyst is too narrow.

My point is that you can be both a developer and a user. And depending on what hat you have on, you have different preferences.

jlstevens (approved):
Looks good to me! 

Thanks for driving this forward @maximlt !

maximlt:
Thanks all for the reviews! With approvals from @droumis, @jbednar, @philippjfr, @hoxbro, @jlstevens, and @ahuang11 (one thumb up somewhere 🙃 ), I think it's fair to consider we've reached a consensus, HEP2 is approved! So next time you intend to deprecate a feature somewhere, please check the HEP :)

(I'll now see with Demetris how to merge it since it depends on https://github.com/holoviz/holoviz/pull/386)

droumis:
I don't think merging this depends on exposing HEPs on the website. It certainly wasn't a blocker for HEP1. Let's go ahead and merge this and then continue the website work.

maximlt:
> I don't think merging this depends on exposing HEPs on the website. It certainly wasn't a blocker for HEP1. Let's go ahead and merge this and then continue the website work.

This PR was branched off https://github.com/holoviz/holoviz/pull/386. Instead, I rebased it on `main`, and "pre-accepted" it by changing the status and adding it to the accepted HEPs file.

> A HoloViz core developer will then formally review the pull request for format, scope, and evidence of consensus. They may also notify other interested parties to weigh in on the HEP. If consensus has been reached, the HoloViz core developer will update the status of the pull request, assign a HEP number, add a link to the discussion in the [HEP table](https://github.com/holoviz/holoviz/blob/main/hep/accepted_heps.md), and merge it. All HEPs will be resolved as either 'rejected', 'accepted', or 'deferred' depending on the consensus of the community.

@droumis, could you have a final look and merge if it's all okay?

PR #390, opened 2024-01-30
Update the contributing guide

maximlt:
- Clarify usage of GitHub service accounts
- Update Analytics with GoatCounter
- Mention 1Password


maximlt:
Self merging since no one did 🙃

PR #392, opened 2024-02-12
Move note about package.json up

ahuang11:


maximlt:
Is this PR superseded by https://github.com/holoviz/holoviz/pull/393?

ahuang11:
Yes, I suppose it is, but depends on which one you want to keep.

PR #396, opened 2024-03-21
Add a custom nav bar with more dropdown menus

maximlt:
@droumis I had a quick look and I think in the end it's not too hard to customize the nav bar to have multiple dropdowns.

image

This PR is not meant to be merged, I think you could use these changes as a basis for your changes to the nav bar (half the relative links in this PR are fake 🙃 ).

I just found the nav bar template for the version we use in this repo (found in the lock file) on the theme's github https://github.com/pydata/pydata-sphinx-theme/blob/v0.13.3/src/pydata_sphinx_theme/theme/pydata_sphinx_theme/components/navbar-nav.html. Saved that in a new file. I inspected the HTML of the nav bar on the currently deployed website, copied and adapted it to add more dropdowns and external links. We'll probably have to update the template when bumping the version of the theme, it shouldn't be too hard though.

---

I added a `Websites` dropdown with links to the HoloViz websites. That's because I've been wondering how to better expose HoloViz and its projects on all the websites. Perhaps that dropdown should be added to all the websites? Just a thought!

---

Since we were wondering, in the newer versions of the theme it's possible to customize the text of the `More` dropdown with the `header_dropdown_text` option (https://github.com/pydata/pydata-sphinx-theme/pull/1423). Useful, but not enough for us I think.

### DIFF
--- a/doc/_templates/nav-bar-holoviz.html
+++ b/doc/_templates/nav-bar-holoviz.html
@@ -1,0 +1,85 @@
+<nav class="navbar-nav">
+  <p class="sidebar-header-items__title"
+     role="heading"
+     aria-level="1"
+     aria-label="{{ _('Site Navigation') }}">
+    {{ _("Site Navigation") }}
+  </p>
+
+  <ul class="bd-navbar-elements navbar-nav">
+    <li class="nav-item">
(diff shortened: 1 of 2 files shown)


droumis:
Nice! thanks. I'll work on the changes to the HoloViz site as we discussed

PR #403, opened 2024-05-18
Update Roadmap.rst

Coderambling:
Changed item 1. for now to:

1.  **More extensive documentation about deployment**: Done. 

Or maybe: **More extensive documentation about deployment**: Update: Done! New extensive documentation can now be found here (links (to Panel deployment)? or somewhere else?).

This would be more informative, and informs users of progress in the roadmap, rather than the item just disappearing from the roadmap, without any info if it has been done or abandoned.

Also, if I remove section 1. entirely, the numbering will start at 2.

I can remove 1. entirely, and re-number the other items (2 through 12), if you confirm the other goals (2 through 12) are still on the roadmap. If more items can be removed, can be removed (or updated to: Done), just let me know and I will edit and re-number accordingly.

Changed Cesium link at bottom to: https://cesium.com/platform/cesiumjs/


jbednar:
Thanks! The current item 2 has been addressed partially, and can be updated to say "Update: Work is ongoing, but several such examples are now at projectpythia.org and https://github.com/holoviz-topics/neuro". Once it has that, I think all the rest are still relevant, so my vote is to remove item 1 entirely, update the current item 2 as indicated, then renumber from the start.

Coderambling:
Ok thanks for the feedback! Please let me know if I'm getting into much detail :-) And maybe versioning of the roadmap on the website should be considered at some point?

Or a more sophisticated roadmap widget that could be re-used for all Holoviz sites :-)

I will try to make the changes as advised, hopefully this weekend.

Coderambling:
Edited the existing fork rather then make a new fork. That's ok right?

jbednar (approved):
Perfect, thanks!

PR #405, opened 2024-05-18
Update background.rst

Coderambling:


doc/background.rst
```diff
@@ -130,7 +130,7 @@ top of lower-level tools, where you can use the top level for anything
 that it provides, while always being able to drop down a level (or
 several if necessary!) to get the behavior you need.
 
-Panel's `interact function <https://panel.pyviz.org/user_guide/Interact.html>`_
+Panel's `interact function <https://panel.holoviz.org/tutorials/intermediate/interactivity.html>`_
```
jbednar, line 133: Thanks for the update! Indeed, the old link was out of date, but the suggested one isn't correct either, since that page doesn't describe the `pn.interact` function that's being referred to here. @philippjfr , up to you how you want to handle references to `interact` like this.

Coderambling: Ah ok. Is the current guidance to use .bind or .rx instead of .interact going forward?

See discussion here: https://github.com/holoviz/panel/issues/6729

Happy to update the doc to whichever link / text.

Also, I see that Panel.pyviz.org/... redirects to Panel.holoviz.org/...

In this case that fails, because that page does not exist, but also because there is no wildcard redirect, which should lead to, for example, the homepage or a dedicated page.

What should happen is:

look for redirect for specific link in the 301 list.

If found, go there.

If not, go to homepage (for now). Or to a dedicated 404 page with search bar.

So a catch-all basically.

I was unable to information on how to configure that in the Rediraffe documentation https://sphinxext-rediraffe.readthedocs.io/en/latest/

PR #407, opened 2024-06-10
For use with Nebari

ahuang11:
`anaconda-project` isn't natively supported in Nebari, meaning users would need to manually `anaconda-project prepare` and that's not good use of tutorial time, so this adds vanilla conda env.yml so that the environment is premade and ready when started.


examples/env.yml
```diff
@@ -0,0 +1,31 @@
+channels:
+  - conda-forge
+dependencies:
+  - python=3.10
+  - pip
+  - pip:
```
ahuang11, line 6: I use `pip` instead of `conda` because Nebari's conda store doesn't really work well with conda. Tried solving these issues, but changing those caused other issues; in end the just went with pip.

image


image

maximlt: Ok noted, it's kind of weird but well it's how it is.

examples/tutorial/00_Setup.ipynb
```diff
@@ -63,8 +65,32 @@
     "    ```\n",
     "    Note: You can replace \"lab\" with \"notebook\" if you prefer the original Jupyter notebook interface.\n",
     "\n",
-    "5. **Test HoloViz imports**:\n",
-    "After launching Jupyter, navigate to this notebook (`tutorial/00_Setup.ipynb`) and run the following cell. This command is a quick check that your environment is set up correctly. You should see the HoloViews, Bokeh, and Matplotlib logos along with a success message.\n"
+    "6. **Test HoloViz imports**:\n",
```
maximlt, line 68: image

To avoid confusing users who follow the recommended path, I'd suggest instead moving the special steps 3, 4, 5 at the end.

examples/tutorial/00_Setup.ipynb
```diff
@@ -63,8 +65,32 @@
...
+    "\n",
+    "3. **Create and activate a new conda environment**: You can also use plain conda to install.\n",
+    "    ```bash\n",
+    "    >> conda env create -f env.yml\n",
+    "    >> conda activate holoviz-tutorial\n",
+    "    ```\n",
+    "4. **Download datasets**: We will be using the following datasets; make sure you're in `examples/` directory before running this.\n",
+    "   ```bash\n",
+    "    mkdir data/raster\n",
+    "    wget -O data/raster/gpw_v4_population_density_rev11_2010_2pt5_min.zip https://earth-data.s3.amazonaws.com/gpw_v4_population_density_rev11_2010_2pt5_min.zip\n",
```
maximlt, line 81: Users will have to unzip this file, anaconda-project does it automatically I think.

examples/tutorial/00_Setup.ipynb
```diff
@@ -63,8 +65,32 @@
...
+    "--- Other Path\n",
+    "\n",
+    "3. **Create and activate a new conda environment**: You can also use plain conda to install.\n",
+    "    ```bash\n",
+    "    >> conda env create -f env.yml\n",
+    "    >> conda activate holoviz-tutorial\n",
+    "    ```\n",
+    "4. **Download datasets**: We will be using the following datasets; make sure you're in `examples/` directory before running this.\n",
+    "   ```bash\n",
+    "    mkdir data/raster\n",
```
maximlt, line 80: On Github at least this section is a bit weirdly formatted 
image

examples/env.yml
maximlt, line 1: To be safer during the tutorial, you might want to pin more precisely the dependencies. But that's up to the people presenting the tutorial really!

jbednar: Yes, it does seem like pinning the versions is much safer.  Can Nebari not just use anaconda-project-lock.yml as its environment file?

examples/tutorial/00_Setup.ipynb
jbednar: Agreed.

jbednar (commented):
Would be nice to get this mergeable.

jbednar (commented):
Would be nice to get this mergeable.

jbednar:
Closing as I think we ended up using conda-store instead.

PR #410, opened 2024-06-24
Update environment and tutorial materials

philippjfr:


ahuang11:
I think we should get this in this week; next week is SciPy. I'll try out your changes soon and update the env on nebari.

examples/anaconda-project.yml
```diff
@@ -6,40 +6,30 @@ maintainers:
...
-- holoviews>=1.16.2
-- pyviz_comms
+- netcdf4
 - plotly
+- pyarrow
+- xarray
 # We don't want to download mkl packages (which are large) mainly seen on Windows
 - nomkl
-# Docs deps, that used to be in a `doc` environment but locking is so slow.
-- nbsite
```
maximlt, line 38: `nbsite` definitely needs to be there to build the website. There used to be two different environments for the tutorial and building the docs (the latter extending the former) but I bundled them as solving was too painfully slow.

PR #412, opened 2024-07-03
Unify place to grab path

ahuang11:
Working on Nebari preparing Scipy tutorial and realized that unless we want all the participants to download the data (on slow conference wifi), we should allow them to easily point the path to the `shared/` data directory.

Alternatively, we can have a bash script to copy data from the `shared/` directory:
`cp -r $HOME/shared/scipy/hvplot-and-panel/* $HOME/tutorials/holoviz/examples/data`

### DIFF
--- a/examples/talks/Overview.ipynb
+++ b/examples/talks/Overview.ipynb
@@ -684,14 +684,19 @@
    },
    "outputs": [],
    "source": [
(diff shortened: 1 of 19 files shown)


jbednar:
Superseded by #413

PR #415, opened 2024-07-08
Update symlink for Scipy shared data directory

ahuang11:
Upon `git clone`, the data directory exists, so no symlinks were created. To address that issue, I loop through each file in the shared data directory to create a symlink for each file.

### DIFF
--- a/examples/tutorial/data_checker.py
+++ b/examples/tutorial/data_checker.py
@@ -14,8 +14,8 @@ def check_data(file_path='../data/earthquakes-projected.parq'):
 
     # Hack to make Nebari installation work for SciPy 2024; can be deleted after the conference
     try:
-        source = Path("/shared/scipy/hvplot-and-panel")
-        dest = Path("../data")
+        source = pathlib.Path("/shared/scipy/hvplot-and-panel")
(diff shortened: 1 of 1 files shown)


jbednar (approved):
Makes sense, thanks!

examples/tutorial/data_checker.py
```diff
@@ -14,12 +14,14 @@ def check_data(file_path='../data/earthquakes-projected.parq'):
...
     # Hack to make Nebari installation work for SciPy 2024; can be deleted after the conference
     try:
-        source = "/shared/scipy/hvplot-and-panel"
-        dest = "../data"
-        if os.path.exists(source) and not os.path.exists(dest):
-            print(f"Adding symlink to data files: {str(dest)} -> {str(source)}")
-            os.symlink(source, dest)
-
+        source = Path("/shared/scipy/hvplot-and-panel")
+        dest = Path("../data")
```
ahuang11:
```suggestion
        source = pathlib.Path("/shared/scipy/hvplot-and-panel")
        dest = pathlib.Path("../data")
```

PR #416, opened 2024-07-08
Reset index

ahuang11:
Without it when trying to use `describe` on selected plot:

```
KeyError: "One or more dimensions in the expression (dim('time')>=np.datetime64('2010-03-11T01:43:21.423041'))&(dim('time')<=np.datetime64('2011-04-01T20:53:28.099845')) could not resolve on ':Dataset   [depth,depthError,dmin,gap,horizontalError,id,latitude,locationSource,longitude,mag,magError,magNst,magSource,magType,net,nst,place,rms,status,type,updated,easting,northing]'. Ensure all dimensions referenced by the expression are present on the supplied object."
```

### DIFF
--- a/examples/tutorial/exercises/Linking_Plots.ipynb
+++ b/examples/tutorial/exercises/Linking_Plots.ipynb
@@ -44,6 +44,7 @@
    "source": [
     "df = pd.read_parquet(pathlib.Path('../../data/earthquakes-projected.parq'))\n",
     "df.index = df.index.tz_localize(None)  # to prevent error in comparison\n",
+    "df = df.reset_index()\n",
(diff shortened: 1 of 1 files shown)


jbednar (approved):
Thanks. No opinion on whether this way or `inplace=True` is clearer.

ahuang11:
`inplace=True` is no longer best practice I think https://github.com/pandas-dev/pandas/issues/16529

jbednar:
@philippjfr , I don't think I should have had to add this line. Can you check it out and decide if a fix is needed in HoloViews, for a datetime axis used as a Pandas index (which appears to be dropped from the HoloViews Dataset being constructed implicitly here)?

PR #417, opened 2024-07-09
Night-before fixes to tutorial for SciPy 2024

jbednar:
Ok, I've finished a complete pass over the entire tutorial, and it technically isn't even the morning of the tutorial yet. As you can see in the diffs, there was a lot of broken stuff! E.g. lots of "solutions" that weren't runnable, and lots of code and explanation scattered around using .interactive rather than .rx (e.g. using variables with "i" instead of "r"). Everything should be working now after this PR, but there are still some to-do items for later:

- [x] 02_Plotting.ipynb: Why only 695 thousand points, rather than the 2.1 million promised? Apparently the data file was recently updated for pyarrow/fastparquet compatibility issues, and somehow a smaller subset was used. Restored the original dataset.
- [ ] 03_Composing_Plots.ipynb last plot: why no colormap for magnitude? I think it's a limitation in HoloViews, but it's one we should address!
- [ ] Exercise linking_plots: "linked_selection aborted because it could not display selection for all elements: Invalid comparison between dtype=datetime64[ns, UTC] and datetime64." (fixed with `df.index = df.index.tz_localize(None)`, but should presumably be fixed at the HoloViews level)
- [ ] Exercise linking_plots: "KeyError: "One or more dimensions in the expression (dim('time')>=np.datetime64('2010-03-11T01:43:21.423041'))&(dim('time')<=np.datetime64('2011-04-01T20:53:28.099845')) could not resolve on ':Dataset   [depth,depthError,dmin,gap,horizontalError,id,latitude,locationSource,longitude,mag,magError,magNst,magSource,magType,net,nst,place,rms,status,type,updated,easting,northing]'. Ensure all dimensions referenced by the expression are present on the supplied object." (fixed with `df = df.reset_index()`, but again presumably HoloViews itself should be fixed?)
- [ ] 05_Interactive_Pipelines.ipynb discusses and explains "terminating methods", but doesn't actually show any, which is deeply confusing. What is that section meant to be any more? Seems like it needs an actual example of termination.
- [ ] 05_Interactive_Pipelines.ipynb filename should be renamed Reactive Expressions, but I didn't want to break things today!
- [ ] Building_a_Dashboard: After a lot of fixes, the suggested solutions now work. But the examples don't seem to lead to anything real -- shouldn't this be an actual working complete dashboard of some sort by the end of it? Instead it's a big non-interactive dashboard followed by a simple interactive example. Confusing!
- [ ] 07_Custom_Interactivity.ipynb, 08_Custom_Dashboards.ipynb, 09_Advanced_Dashboards.ipynb: I didn't have time to run the inline exercises in these last three notebooks, so I haven't checked that the hints are still ok. Unlike the earlier ones, there are no solutions, or at least nothing labeled solutions (some hints do seem to be solutions, so these should get updated.
- [ ] The .py files in the tutorial all seem out of date, and need to be updated to match the contents of the .ipynb files (or be deleted)


jbednar:
Sorry about the metadata -- depending on the file, stripping it out caused diffs, and on others just clearing the notebook caused diffs, so in the end I just treated them all the same, and any reviewer will have to just ignore the notebook metadata.

examples/tutorial/03_Composing_Plots.ipynb
```diff
@@ -158,7 +158,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The two timeseries have quite different value ranges, making it very hard to see the fluctuations in magnitude with an overlay like this. It is possible to zoom in to see the detail in `mag`, but a more useful form of composition here is a layout of separate plots, using a `+` symbol to compose HoloViews objects side-by-side with axes linked for any shared dimensions:"
+    "The two timeseries have quite different value ranges, making it very hard to see the fluctuations in magnitude with an overlay like this. It is possible to zoom in to see the detail in `mag`, or you can try multiple y axes using syntax we will discuss later `(weekly_mean_magnitude_plot * weekly_count_plot).opts(multi_y=True)`). A more useful form of composition here is a layout of separate plots, using a `+` symbol to compose HoloViews objects side-by-side with axes linked for any shared dimensions:"
```
jbednar:
```suggestion
    "The two timeseries have quite different value ranges, making it very hard to see the fluctuations in magnitude with an overlay like this. It is possible to zoom in to see the detail in `mag`, or you can try multiple y axes using syntax we will discuss later `(weekly_mean_magnitude_plot * weekly_count_plot).opts(multi_y=True)`. A more useful form of composition here is a layout of separate plots, using a `+` symbol to compose HoloViews objects side-by-side with axes linked for any shared dimensions:"
```

PR #424, opened 2024-11-14
Automatically push tutorial archive to S3 and update setup instructions

maximlt:
`anaconda-project download` no longer works (feature removed from anaconda.org), so we needed to find a new approach. In this PR, the tutorial is automatically updated to S3 (on release or workflow_dispatch of the docs workflow targeting the main site), and the install instructions are updated accordingly.


examples/tutorial/00_Setup.ipynb
```diff
@@ -51,9 +51,21 @@
...
+    "    Expand-Archive -Path \"holoviz_tutorial.zip\" -DestinationPath \"holoviz_tutorial\"\n",
+    "    cd holoviz_tutorial\n",
+    "    ```\n",
+    "\n",
+    "    *macOS and Linux*\n",
+    "\n",
     "    ```bash\n",
-    "    anaconda-project download pyviz/holoviz_tutorial\n",
+    "    curl -o holoviz_tutorial.zip \"https://assets.holoviz.org/holoviz/tutorial/holoviz_tutorial.zip\"\n",
+    "    unzip holoviz_tutorial.zip -d holoviz_tutorial\n",
```
maximlt, line 68: Used ChatGPT to generate these install instructions so I believe they're generic enough.

.github/workflows/docs.yaml
```diff
@@ -66,7 +69,9 @@ jobs:
         run: git fetch --prune --tags --unshallow -f
       - name: Set output
         id: vars
-        run: echo "tag=${GITHUB_REF#refs/*/}" >> $GITHUB_OUTPUT
+        run: |
+          echo "tag=${GITHUB_REF#refs/*/}" >> $GITHUB_OUTPUT
+          echo "fdatetime=$(date +%Y%m%d-%H%M)" >> $GITHUB_OUTPUT
       - name: install anaconda-project
         run: conda install "anaconda-project>=0.10.1"
```
droumis, line 76: The instructions in 00_Setup.ipynb has `anaconda-project>=0.11` but the docs.yaml uses `anaconda-project>=0.10.1`... is there a reason?

droumis (approved):
All looks good. 

The minor anaconda-project version mismatch is not even a part of this PR change so feel free to ignore or fix.

.github/workflows/docs.yaml
maximlt: Just that one was updated and not the other :)

PR #429, opened 2025-03-18
Amend HEP 1: Dropping Python version support must NOT be done in a patch release

maximlt:
Since this is effectively what we (try to!) do.


droumis:
Do you think we need to add something to HEP 0 about the process for amendments?

maximlt:
I guess ideally yes? HEP 0 shows the metadata of a HEP should have an `Updated` field but doesn't explain when a HEP could be updated and how.

droumis:
@holoviz/holoviz-dev

droumis:
Policy to amend HEPs: https://github.com/holoviz/holoviz/pull/434

droumis (approved):

jbednar (approved):

maximlt:
Thanks!

PR #439, opened 2026-01-29
Add lumen logo

ahuang11:


jbednar (approved):

philippjfr:
Not sure yellow text is great for light backgrounds.

ahuang11:
Since it's dark yellow, it looks fine to me.

image
image

Also, yellow is the color I think of when I think about lighbulbs (lumen), plus I wanted a unique hue from the other packages.
image

PR #442, opened 2026-04-03
Remove Sophia Yang from Steering Committee

droumis:
Sophia Yang has resigned from the HoloViz Steering Committee, per Section 4.1 of the Charter. 

Her resignation was communicated to Jim Bednar in writing and verbally relayed to the Steering Committee at the March 27, 2026 meeting.

This PR removes her entry from the Steering Committee membership list.

Requesting review from @sophiamyang or @jbednar to confirm.


jbednar (approved):
Approved, albeit with deep regrets!  @sophiamyang , you will be missed.

PR #451, opened 2026-06-28
Moved tutorial zip files into holoviz_tutorial subdirectory

jbednar:
Moved tutorial zip files into holoviz_tutorial subdirectory for isolation from homedir. 

With the previous version, when I unpacked it into my home directory, pixi was recursively scanning the many thousands of files in my homedir and reporting a bunch of irrelevant issues like dangling symlinks. 

Updated the Setup docs to match.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -87,7 +87,7 @@ jobs:
       - name: git clean
         run: git clean -fxd .
       - name: zip tutorial
-        run: mv examples holoviz_tutorial && zip -r holoviz_tutorial3.zip holoviz_tutorial -x "holoviz_tutorial/data/*" ; mv holoviz_tutorial examples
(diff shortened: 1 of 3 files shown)


Azaya89:
Made a small improvement with the setup: Someone who just downloads the zip files and runs the tutorial doesn't need linting since they're not making any dev changes.

Azaya89 (approved):

jbednar:
I think @maximlt had reported that this change wasn't necessary, i.e. that when he unpacked the earlier version of the tutorial, not embedded in a subdirectory, he had no problems. Interestingly, I just replicated that on my own machine, a Macbook: if I double-click on the zip file, it gets unpacked into a subdirectory, whereas if I do `unzip holoviz_tutorial.zip` on the command line, it gets unpacked in place (causing problems in the home dir).  

After this change, if someone double-clicks on it on a Mac, they'll get a directory structure like holoviz_tutorial/holoviz_tutorial/tutorial, but that still seems safer than the alternative.

